PR #7, opened 2024-05-15
Fix version to 0.0.1

gclssvglx:
# Set version to 0.0.1

## JIRA ticket(s)
  - [GOVAPP-473](https://govukverify.atlassian.net/browse/GOVAPP-473)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -26,7 +26,7 @@ android {
         minSdk = 24
         targetSdk = 34
         versionCode = 1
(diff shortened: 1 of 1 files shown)


batesyuk (approved):
No sonar in this one?

gclssvglx:
> No sonar in this one?

No, I figured there was little point since the modules are already tested, scanned etc and the actual integration should be minimal I think. Happy to add it if you'd prefer.

PR #54, opened 2024-08-28
GOVUKAPP-669 Add alt text to Homepage logo

gclssvglx:
# Add alt text to Homepage logo

## JIRA ticket(s)
  - [GOVAPP-669](https://govukverify.atlassian.net/browse/GOVUKAPP-669)

### DIFF
--- a/feature/home/src/main/kotlin/uk/govuk/app/home/ui/HomeScreen.kt
+++ b/feature/home/src/main/kotlin/uk/govuk/app/home/ui/HomeScreen.kt
@@ -134,7 +134,7 @@ private fun ScalingHeader(
     Column(modifier = modifier) {
         Image(
             painter = painterResource(id = uk.govuk.app.design.R.drawable.logo),
-            contentDescription = null,
+            contentDescription = "GOV.UK Logo",
             modifier = Modifier
                 .padding(vertical = padding.dp)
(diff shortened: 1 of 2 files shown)


feature/home/src/main/kotlin/uk/govuk/app/home/ui/HomeScreen.kt
```diff
@@ -134,7 +134,7 @@ private fun ScalingHeader(
     Column(modifier = modifier) {
         Image(
             painter = painterResource(id = uk.govuk.app.design.R.drawable.logo),
-            contentDescription = null,
+            contentDescription = "GOV.UK Logo",
```
TimGJD, line 137: Wondering if this might need going in a string resource file maybe ready for Welsh translation of logo? (not sure what logo is in Welsh?)

gclssvglx: Happy to add this as a string resource. However, according to Google translate at least - the Welsh for `Logo` is `Logo`, though it is pronounced differently to the English version.

As an aside - when read by TalkBack - it appends the word `image` to that phrase. So, we get `GOV.UK Logo image`. This seems build-in and so not changeable. One for Post Amigo I think.

TimGJD (approved):

PR #71, opened 2024-09-25
GOVUKAPP-760: R8 crash fix

davidc-gds:
# Title

Add rules to ensure R8 (shrinking and obfuscation) does not cause release (alpha) builds to crash

## JIRA ticket(s)
  - [GOVUKAPP-760](https://govukverify.atlassian.net/browse/GOVUKAPP-760)


gclssvglx (approved):
I'm not sure I totally get this, but I get that un-serialized data won't work correctly and this looks like it's fixing that. Great bit of sleuthing 🔍

PR #76, opened 2024-10-02
GOVUKAPP-586 verify config signature

gclssvglx:
# GOVUKAPP-586 verify config signature

Implement a mechanism within the mobile app to verify the signature of the configuration data using the bundled public key.

## JIRA ticket(s)
  - [GOVAPP-586](https://govukverify.atlassian.net/browse/GOVUKAPP-586)

### DIFF
--- a/config/src/main/kotlin/uk/govuk/app/config/ConfigRepo.kt
+++ b/config/src/main/kotlin/uk/govuk/app/config/ConfigRepo.kt
@@ -9,7 +9,9 @@ import javax.inject.Singleton
 
 @Singleton
 class ConfigRepo @Inject constructor(
-    private val configApi: ConfigApi
+    private val configApi: ConfigApi,
+    private val gson: Gson,
+    private val signatureValidator: SignatureValidator
 ) {
     private lateinit var _config: Config
     val config: Config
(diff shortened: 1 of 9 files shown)


config/src/main/kotlin/uk/govuk/app/config/ConfigRepo.kt
```diff
@@ -24,14 +28,22 @@ class ConfigRepo @Inject constructor(
...
+
+                    _config = configResponse.config
                     true
                 } ?: false
             } else {
                 false
             }
         } catch (e: Exception) {
-            false
+            return false
```
davidc-gds, line 46: Don't think we need the return keyword here.

config/src/main/kotlin/uk/govuk/app/config/SignatureValidator.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Base64
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class SignatureValidator @Inject constructor() {
+    private val key = BuildConfig.CONFIG_PUBLIC_KEY
+
+    fun isValidSignature(
+        signature: String = "",
```
davidc-gds, line 17: Don't think we need the default values.

davidc-gds (approved):

PR #81, opened 2024-10-11
GOVUKAPP-679: display open source licences

gclssvglx:
# GOVUKAPP-679: display open source licences

We need a solution to display the app's dependency licences - in-app - with relevant developer information and outward links, which doesn’t require manual population or administration.

Added as per the instructions here:

- https://github.com/google/play-services-plugins/tree/main/oss-licenses-plugin
- https://developers.google.com/android/guides/opensource

## JIRA ticket(s)
  - [GOVAPP-679](https://govukverify.atlassian.net/browse/GOVUKAPP-679)

## Figma
  - [Figma board](https://www.figma.com/design/5u5cvVCgKY3Rng2ADNV30K/2024-07---GOV.UK-app---beta-designs?node-id=1-82977&node-type=canvas&t=bE8WW67KarKR6IY8-0)

## Screenshots

| Light Mode | Dark Mode |
| --- | --- |
| light-screen-1 | dark-screen-1 |
| light-screen-2 | dark-screen-2 |

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     alias(libs.plugins.androidApplication)
+    id("com.google.android.gms.oss-licenses-plugin")
     alias(libs.plugins.jetbrainsKotlinAndroid)
     alias(libs.plugins.compose)
     alias(libs.plugins.hilt)
@@ -104,6 +105,7 @@ dependencies {
     implementation(libs.androidx.datastore.preferences)
 
     implementation(libs.lottie.compose)
+    implementation(libs.play.services.oss.licenses)
 
     ksp(libs.hilt.compiler)
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -15,11 +15,8 @@ plugins {
 }
 
 buildscript {
-    repositories {
-        google()
-    }
     dependencies {
-        classpath(libs.oss.licenses.plugin)
+        classpath(libs.oss.licenses)
     }
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,6 +14,15 @@ plugins {
     alias(libs.plugins.sonarqube)
 }
 
+buildscript {
+    repositories {
+        google()
+    }
+    dependencies {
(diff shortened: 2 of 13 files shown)


app/build.gradle.kts
```diff
@@ -1,5 +1,6 @@
 plugins {
     alias(libs.plugins.androidApplication)
+    id("com.google.android.gms.oss-licenses-plugin")
```
davidc-gds, line 3: Can we move this into the libs file?

build.gradle.kts
```diff
@@ -14,6 +14,15 @@ plugins {
     alias(libs.plugins.sonarqube)
 }
 
+buildscript {
+    repositories {
```
davidc-gds, line 18: We may not need this here as the repositories are already defined in settings.gradle

feature/settings/src/main/kotlin/uk/govuk/app/settings/ui/SettingsScreen.kt
```diff
@@ -235,3 +245,45 @@ private fun PrivacyAndLegal(
...
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(GovUkTheme.spacing.medium)
+        ) {
+            Row(
+                Modifier
+                    .clickable(onClick = {
+                        onLicenseView()
+                        context.startActivity(
+                            Intent(
```
davidc-gds, line 272: Can we move this into SettingsNavigation (similar to privacy policy)?

app/build.gradle.kts
gclssvglx: Umm, doesn't seem to work. So, the dependency is already in `gradle/libs.versions.toml` lines 19 & 61. However, when I try to remove this from here, the plugin then reports that the app doesn't have any open source licenses...

no-license-message

gclssvglx: Oh, IC - sorry.

davidc-gds: We can look into this in the future, it's not a big issue.

davidc-gds (approved):

PR #91, opened 2024-10-22
GOVUKAPP-649 : visited item list storage

gclssvglx:
# Vsited item list storage

List Storage, including:
- Create storage
- Save service view to storage
- Fetch from storage

## JIRA ticket(s)
  - [GOVAPP-649](https://govukverify.atlassian.net/browse/GOVUKAPP-649)

### DIFF
--- a/feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedRepo.kt
+++ b/feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedRepo.kt
@@ -1,0 +1,12 @@
+package uk.govuk.app.visited.data
+
+import uk.govuk.app.visited.data.store.VisitedLocalDataSource
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+internal class VisitedRepo @Inject constructor(
+    localDataSource: VisitedLocalDataSource,
+) {
+    val visitedItems = localDataSource.visitedItems
+}
--- a/feature/visited/src/main/kotlin/uk/govuk/app/visited/data/model/VisitedItem.kt
+++ b/feature/visited/src/main/kotlin/uk/govuk/app/visited/data/model/VisitedItem.kt
@@ -1,0 +1,12 @@
+package uk.govuk.app.visited.data.model
+
+import io.realm.kotlin.types.RealmObject
+import io.realm.kotlin.types.annotations.PrimaryKey
+import org.mongodb.kbson.ObjectId
+
+internal class VisitedItem: RealmObject {
+    @PrimaryKey var _id: ObjectId = ObjectId()
+    var title: String = ""
+    var url: String = ""
+    var lastVisited: String = "" // TODO: needs to be a date at this point
+}
--- a/feature/visited/src/main/kotlin/uk/govuk/app/visited/ui/VisitedScreen.kt
+++ b/feature/visited/src/main/kotlin/uk/govuk/app/visited/ui/VisitedScreen.kt
@@ -1,25 +1,37 @@
 package uk.govuk.app.visited.ui
 
 import android.annotation.SuppressLint
+import android.content.Intent
+import android.net.Uri
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
+import androidx.compose.material3.CardDefaults
+import androidx.compose.material3.Icon
+import androidx.compose.material3.OutlinedCard
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
 import uk.govuk.app.design.ui.component.BodyBoldLabel
 import uk.govuk.app.design.ui.component.BodyRegularLabel
 import uk.govuk.app.design.ui.component.ChildPageHeader
 import uk.govuk.app.design.ui.component.ExtraLargeVerticalSpacer
+import uk.govuk.app.design.ui.component.SubheadlineRegularLabel
 import uk.govuk.app.design.ui.theme.GovUkTheme
 import uk.govuk.app.visited.R
 import uk.govuk.app.visited.VisitedViewModel
+import uk.govuk.app.visited.ui.model.VisitedUi
 
 @Composable
 internal fun VisitedRoute(
@@ -69,7 +81,7 @@ private fun VisitedScreen(
                 if (visitedItems.isEmpty()) {
                     NoVisitedItems(modifier)
                 } else {
-                    /* TODO: display the visited items */
+                    ShowVisitedItems(visitedItems, viewModel, modifier)
                 }
             }
         }
@@ -95,5 +107,78 @@ private fun NoVisitedItems(
             stringResource(R.string.visited_items_no_pages_description),
             textAlign = TextAlign.Center
         )
+    }
+}
+
+@Composable
+private fun ShowVisitedItems(
+    items: List<VisitedUi>?,
+    viewModel: VisitedViewModel,
+    modifier: Modifier = Modifier
+) {
+    val lastVisitedText = stringResource(R.string.visited_items_last_visited)
+
+    items?.forEach { item ->
+        val title = item.title
+        val lastVisited = item.lastVisited
+        val url = item.url
+        val context = LocalContext.current
+        val intent = Intent(Intent.ACTION_VIEW)
+        intent.data = Uri.parse(url)
+
+        OutlinedCard(
+            colors = CardDefaults.cardColors(
+                containerColor = GovUkTheme.colourScheme.surfaces.card
+            ),
+            modifier = modifier.padding(
+                GovUkTheme.spacing.medium,
+                GovUkTheme.spacing.medium,
+                GovUkTheme.spacing.medium,
+                0.dp
+            )
+                .fillMaxWidth()
+                .clickable(
+                    onClick = {
+                        viewModel.onVisitedItemClicked(title, url)
+                        context.startActivity(intent)
+                    }
+                ),
+        ) {
+            Row(
+                modifier.padding(
+                    GovUkTheme.spacing.medium,
+                    GovUkTheme.spacing.medium,
+                    GovUkTheme.spacing.medium,
+                    GovUkTheme.spacing.small
+                ),
+                verticalAlignment = Alignment.Top
+            ) {
+                BodyRegularLabel(
+                    text = title,
+                    modifier = Modifier.weight(1f),
+                    color = GovUkTheme.colourScheme.textAndIcons.link,
+                )
+
+                Icon(
+                    painter = painterResource(
+                        uk.govuk.app.design.R.drawable.baseline_open_in_new_24
+                    ),
+                    contentDescription = stringResource(
+                        uk.govuk.app.design.R.string.opens_in_web_browser
+                    ),
+                    tint = GovUkTheme.colourScheme.textAndIcons.link
+                )
+            }
+
+            SubheadlineRegularLabel(
+                text = "$lastVisitedText $lastVisited",
+                modifier = modifier.padding(
+                    GovUkTheme.spacing.medium,
+                    0.dp,
+                    GovUkTheme.spacing.medium,
+                    GovUkTheme.spacing.medium
+                )
+            )
+        }
     }
 }
--- a/feature/visited/build.gradle.kts
+++ b/feature/visited/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
     alias(libs.plugins.hilt)
     alias(libs.plugins.ksp)
     alias(libs.plugins.kover)
+    alias(libs.plugins.realm)
 }
 
 android {
@@ -35,6 +36,8 @@ dependencies {
     implementation(libs.androidx.material3)
     implementation(libs.androidx.hilt.navigation.compose)
     implementation(libs.hilt.android)
+    implementation(libs.androidx.datastore.preferences)
+    implementation(libs.realm.base)
 
     ksp(libs.hilt.compiler)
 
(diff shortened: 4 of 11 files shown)


davidc-gds (dismissed):

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
```diff
@@ -122,6 +124,7 @@ private fun SearchScreen(
 @Composable
 fun ShowResults(searchResults: List<Result>) {
     val viewModel: SearchViewModel = hiltViewModel()
+    val scope = rememberCoroutineScope()
```
davidc-gds, line 127: This isn't required, the view model function can launch a view model scope block.

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
```diff
@@ -151,6 +154,9 @@ fun ShowResults(searchResults: List<Result>) {
                         .fillMaxWidth()
                         .clickable(
                             onClick = {
+                                scope.launch {
```
davidc-gds, line 157: The visited item functionality could be moved into the function that handles the search result click.

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
```diff
@@ -122,6 +124,7 @@ private fun SearchScreen(
 @Composable
 fun ShowResults(searchResults: List<Result>) {
     val viewModel: SearchViewModel = hiltViewModel()
```
davidc-gds, line 126: Better to pass lambdas to the composables from the route rather than pass the view model or access it directly.

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    init {
+        todaysItems = todaysItems()
+        thisMonthsItems = thisMonthsItems()
+        previousMonthsItems = previousMonthsItems()
+        groupedPreviousMonthsItems = groupPreviousMonthsItems()
+    }
+
+    fun transform(): Map<String, List<VisitedUi>> {
+        var groupedVisitedItems = emptyMap<String, List<VisitedUi>>()
```
davidc-gds, line 24: This should be a mutable map as concatenation of immutable map creates lots of redundant objects.

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/model/VisitedItem.kt
```diff
@@ -0,0 +1,12 @@
+package uk.govuk.app.visited.data.model
+
+import io.realm.kotlin.types.RealmObject
+import io.realm.kotlin.types.annotations.PrimaryKey
+import org.mongodb.kbson.ObjectId
+
+class VisitedItem: RealmObject {
+    @PrimaryKey var _id: ObjectId = ObjectId()
```
davidc-gds, line 8: Is an id required, can we not just use the url as the primary key?

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
```diff
@@ -0,0 +1,78 @@
...
+
+        if (todaysItems.isNotEmpty()) {
+            groupedVisitedItems += mapOf("Today" to toVisitedUi(todaysItems))
```
davidc-gds, line 27: Are the keys visible on the uI?

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
```diff
@@ -0,0 +1,78 @@
+package uk.govuk.app.visited.data
+
+import uk.govuk.app.visited.data.model.VisitedItem
+import uk.govuk.app.visited.ui.model.VisitedUi
+import java.time.LocalDate
+
+class VisitedItemsTransformer(
```
davidc-gds, line 7: This could probably be a stand alone (or extension) function rather than a class.

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedRepo.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import uk.govuk.app.visited.data.store.VisitedLocalDataSource
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+internal class VisitedRepo @Inject constructor(
+    localDataSource: VisitedLocalDataSource,
+) {
+    val visitedItems = localDataSource.visitedItems
```
davidc-gds, line 11: It would probably be better to map the realm object to a domain object to ensure encapsulation of realm within the data layer.

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
gclssvglx: Yes, so prob should be string values - will fix

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/model/VisitedItem.kt
gclssvglx: This is an interesting question!

Basically, we could use the URL - but they are not unique over time (like most 'business' keys) and that isn't going to help the user or us at all, and it won't stop multiple entries pointing to the same page. So, while they look like a good option, they really aren't.

Back story... 

We're using (effectively) a content item. These have a title, a content ID, a document type and a path. The path is built from the title, plus the document type - both of which can be changed by the user. This means the path can also change (triggered by the user). They have 3 options: leave the path as-is, update it based on the new title and/or document type, or redirect the old page to the new one. If we were using the content ID - which is unique and unchanging for the entire lifecycle of the content item - then that would be a clear preference for the key here.

So, my thinking is - if we can't find a useful 'business' key - then we shouldn't use one as it'll send the wrong message to other devs on the app.

davidc-gds: So this means we could get multiple entries for the same url but some of them will have title's that don't match what is currently on the page?

gclssvglx: Yes, it is possible.

davidc-gds: What would happen if you access the same page multiple times (not from the visited pages screen)? I'm thinking you would get multiple entries, is this the desired behaviour?

gclssvglx: That wouldn't be desirable no. However, when we `insertOrUpdate` we check for existing visited items by title and url - should stop this happening I think.

gclssvglx: So, we don't use the id for this reason.

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
gclssvglx: Fixed up

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
gclssvglx: Fixed up

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
gclssvglx: Fixed up

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
gclssvglx: Fixed up

gclssvglx: Solution is not great, but at least there's only one place to update now.

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
gclssvglx: Fixed up

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedRepo.kt
gclssvglx: Fixed up

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
gclssvglx: Fixed up

feature/search/src/main/kotlin/uk/govuk/app/search/SearchViewModel.kt
```diff
@@ -58,8 +60,9 @@ internal class SearchViewModel @Inject constructor(
         analytics.search(searchTerm)
     }
 
-    fun onSearchResultClicked(title: String, url: String) {
+    suspend fun onSearchResultClicked(title: String, url: String) {
```
davidc-gds, line 63: This doesn't need to be a suspend function, it can launch a coroutine via the view model scope.

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
```diff
@@ -60,6 +62,11 @@ internal fun SearchRoute(
         onClear = {
             viewModel.onClear()
         },
+        onClick = { title, url ->
+            viewModel.viewModelScope.launch {
```
davidc-gds, line 66: The view model scope should be launched within the view model function.

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
```diff
@@ -100,7 +108,7 @@ private fun SearchScreen(
 
         when (uiState?.resultStatus) {
             ResultStatus.Success ->
-                ShowResults(uiState!!.searchResults)
+                ShowResults(uiState!!.searchResults, onClick)
```
davidc-gds, line 111: We should avoid using !! if possible, something like this would work:

`uiState?.let {
            when (it.resultStatus) {
                ResultStatus.Success ->
                    ShowResults(it.searchResults)
                ResultStatus.Empty ->
                    NoResultsFound(searchTerm = it.searchTerm)
                ResultStatus.DeviceOffline ->
                    DeviceIsOffline()
                ResultStatus.ServiceNotResponding ->
                    ServiceNotResponding()
                else ->
                    ShowNothing()
            }
        } ?: ShowNothing()`

feature/visited/src/main/kotlin/uk/govuk/app/visited/ui/VisitedScreen.kt
```diff
@@ -97,3 +117,70 @@ private fun NoVisitedItems(
         )
     }
 }
+
+@Composable
+private fun ShowVisitedItems(
+    items: Map<String, List<VisitedUi>>?,
```
davidc-gds, line 123: Not sure this needs to be nullable?

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
davidc-gds: I'm still seeing it as a class

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
davidc-gds: Hard coded strings still appear to be present

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
gclssvglx: That's odd 🤔 You don't see `fun transformVisitedItems(visitedItems: List<VisitedItemUi>, todaysDate: LocalDate): Map<String, List<VisitedUi>> {` [here](https://github.com/alphagov/govuk-mobile-android-app/pull/91/commits/8a221f34054365ebf858b17937c2e26f1c3eab51#diff-40c77ccd20d05237a073377c0cf1c4f5eb18ed2b5fa79bea8dc14397f3c45295)

feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
gclssvglx: That's odd 🤔 They should have been replaced via this commit: https://github.com/alphagov/govuk-mobile-android-app/pull/91/commits/1767724e4aa865643fea8d089635bc933be4c104

gclssvglx: Admittedly, there are still strings in the SectionTitles file

gclssvglx: I'll try to replace them with @StringRef

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
gclssvglx: Fixed up

feature/visited/src/main/kotlin/uk/govuk/app/visited/ui/VisitedScreen.kt
gclssvglx: Fixed up

feature/search/src/main/kotlin/uk/govuk/app/search/ui/SearchScreen.kt
gclssvglx: Fixed up

feature/search/src/main/kotlin/uk/govuk/app/search/SearchViewModel.kt
gclssvglx: Fixed up

davidc-gds (approved):

PR #92, opened 2024-10-25
GOVUKAPP-808: Topic page

davidc-gds:
# Title

Topic page

## JIRA ticket(s)
  - [GOVUKAPP-808](https://govukverify.atlassian.net/browse/GOVUKAPP-808)

## Figma
  - [Figma board](https://www.figma.com/design/5u5cvVCgKY3Rng2ADNV30K/2024-07---GOV.UK-app---beta-designs?node-id=338-13767&t=lOWAMpZeGdpPdLDi-4)

## Screen shots & video (if UI changes)
Screenshot_1729865693

[GOVUKAPP-808]: https://govukverify.atlassian.net/browse/GOVUKAPP-808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


feature/topics/src/main/kotlin/uk/govuk/app/topics/ui/TopicScreen.kt
```diff
@@ -1,27 +1,226 @@
...
+                    stepBySteps = topic.stepBySteps,
+                    displaySeeAll = topic.displayStepByStepSeeAll,
+                    onClick = onExternalLink,
+                    onSeeAll = onStepByStepSeeAll
+                )
+
+                subtopics(
+                    subtopics = topic.subtopics,
+                    onClick = onSubtopic
+                )
```
gclssvglx, line 114: Maybe not a question for you or here. But, I'm wondering what, if anything, are we doing about cyclic loops in the Topics hierarchy 🤔  While clicking through the UI at random I noticed that I was being led from one top level, into another top-level's sub-topics structure. I found this oddly disconcerting as it felt like I'd literally gone off-topic 😕 I imagine some of our users could get very confused by this.

feature/topics/src/main/kotlin/uk/govuk/app/topics/ui/widget/TopicsWidget.kt
```diff
@@ -187,39 +187,39 @@ private fun TopicsWidgetPreview() {
     GovUkTheme {
         TopicsWidgetContent(
             topics = listOf(
-                TopicUi(
+                TopicItemUi(
                     "",
                     uk.govuk.app.design.R.drawable.ic_topic_default,
                     "A really really really really really really long topic title",
```
gclssvglx, line 193: It's a great idea to test this! But, we might even longer text - see screenshot where the description has completely disappeared and the title is chopped. There may well be even longer ones waiting in the wings too as there's no char limit as far as I know on this field in the content API. Could/should we chop it and give it an ellipsis, e.g. "Being diagnosed with a ...", shrink the font (I thinking this is not easy or desirable for accessibility), something else?

I'm guessing this design issue needs to be addressed, so you're waiting for a steer ❓ 

Screenshot_20241028_085540

gclssvglx (approved):
Fantastic stuff @davidc-gds 🥇 

No blockers for this PR, but a couple of questions that may need addressing at some point.

design/src/main/kotlin/uk/govuk/app/design/ui/component/Button.kt
```diff
@@ -198,14 +198,14 @@ private fun BaseButton(
...
     }
 }
 
 @Composable
-private fun NewTabIcon() {
+private fun ExternalLinkIcon() {
     Icon(
-        painter = painterResource(id = R.drawable.baseline_open_in_new_24),
+        painter = painterResource(id = R.drawable.ic_external_link),
         contentDescription = "Opens link in a new tab",
```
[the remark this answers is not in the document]
davidc-gds: Yes it should, it should probably also use the same "opens in web browser" string that the other external links use.

feature/topics/src/main/kotlin/uk/govuk/app/topics/ui/TopicScreen.kt
davidc-gds: Yeh, I noticed that you can get pretty deep into it. Not sure if it's been considered, I'll raise it with Ryan.

feature/topics/src/main/kotlin/uk/govuk/app/topics/ui/widget/TopicsWidget.kt
davidc-gds: I was going to raise this in post amigo. It's mostly an issue with the header being a fixed height and also the design not taking titles that wrap into account. It looks rubbish when it wraps onto two lines even if text isn't getting cut off as the gap between the title and the content is too small (issue with using a fixed height rather than padding).

design/src/main/kotlin/uk/govuk/app/design/ui/component/Button.kt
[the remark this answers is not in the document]
davidc-gds: Good spot!

feature/topics/src/main/kotlin/uk/govuk/app/topics/ui/widget/TopicsWidget.kt
gclssvglx: Yeah, I figured you might let Post Amigo deal with this design issue - I think that's a wise move. Definitely need some guidance on this. It'll effect visited items and potentially other areas of the app too 👍

feature/topics/src/main/kotlin/uk/govuk/app/topics/ui/TopicScreen.kt
davidc-gds: Just raised it in post amigo, they are aware of it as a known risk, it's down to the content team to manage it. On iOS they have a work around where if you hit the home tab it boots you back up to the home page. Will probably implement the same thing on Android.

gclssvglx: I don't think this is something we can or should try to solve within the app. I feel it needs to be resolved at the Topic hierarchy level.

design/src/main/kotlin/uk/govuk/app/design/ui/component/Button.kt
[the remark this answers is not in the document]
davidc-gds: Fixed

PR #105, opened 2024-11-07
Switch Search API to V2

gclssvglx:
# Switch Search API to V2

- [Search v1](https://www.gov.uk/api/search.json?q=micropig)
- [Search v2](https://search.publishing.service.gov.uk/v0_1/search.json?q=micropig)

## Notes 🎶 

- `description` in v1 has been renamed/moved to `description_with_highlighting`
- `description_with_highlighting` is an optional field

## JIRA ticket(s)
  - [GOVAPP-833](https://govukverify.atlassian.net/browse/GOVUKAPP-833)

### DIFF
--- a/feature/search/src/main/kotlin/uk/govuk/app/search/data/remote/model/Result.kt
+++ b/feature/search/src/main/kotlin/uk/govuk/app/search/data/remote/model/Result.kt
@@ -1,9 +1,10 @@
 package uk.govuk.app.search.data.remote.model
 
 import com.google.gson.annotations.SerializedName
+import uk.govuk.app.search.domain.SearchConfig.DESCRIPTION_RESPONSE_FIELD
 
 data class Result(
     @SerializedName("title") val title: String,
-    @SerializedName("description_with_highlighting") val description: String?,
+    @SerializedName(DESCRIPTION_RESPONSE_FIELD) val description: String?,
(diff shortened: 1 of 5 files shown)


feature/search/src/main/kotlin/uk/govuk/app/search/data/remote/model/Result.kt
```diff
@@ -4,6 +4,6 @@ import com.google.gson.annotations.SerializedName
 
 data class Result(
     @SerializedName("title") val title: String,
-    @SerializedName("description") val description: String,
+    @SerializedName("description_with_highlighting") val description: String?,
```
davidc-gds, line 7: Is this going to cause an issue if we have to switch back to v1?

gclssvglx: Yup, `description` simply doesn't exist. I could make this `DESCRIPTION` and map it in the SearchConfig - it'll help a bit - but fundamentally if we're switching back it'll not be as quick and easy as it should have been

gclssvglx: I've added another commit to do as suggested above

davidc-gds (approved):

PR #106, opened 2024-11-08
GOVUKAPP-590 Global ON/OFF switch

TimGJD:
# Title
Global ON / OFF switch

# Description
Add app unavailable screen and the logic to show it

## JIRA ticket(s)
  - [GOVAPP-590](https://govukverify.atlassian.net/browse/GOVUKAPP-590)

## Figma
  - [Figma board](https://www.figma.com/design/5u5cvVCgKY3Rng2ADNV30K/2024-07---GOV.UK-app---beta-designs?node-id=363-6730&node-type=section&t=ZK9eNXBoSekBxAnK-0)

### After

| Light Mode | Dark Mode |
|---|---|
| image | image |


feature/availability/src/main/kotlin/uk/govuk/app/availability/ui/UnavailableScreen.kt
```diff
@@ -0,0 +1,82 @@
...
+private fun GoToGovUkButton(
+    onGoToGovUkClick: () -> Unit
+) {
+    val text = stringResource(R.string.unavailable_button_title)
+    val altText = stringResource(R.string.unavailable_button_title_alt_text)
+    SecondaryButton(
+        text = text,
+        onClick = { onGoToGovUkClick() },
+        modifier = Modifier
+            .clearAndSetSemantics { contentDescription = altText },
```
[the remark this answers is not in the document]
gclssvglx: Ah, `govenor` is a known issue with talkback. 

If it's visible to the user, then I believe we've accepted that this might happen and have chosen to proceed with it rather than try to fix it everywhere. 

If `GOV.UK` is not visible - just read by talkback - it's ok to replace `GOV.UK` with `GOV UK` (space, not dot).

gclssvglx (dismissed):
This is looking good to me 👍 

However, I'd wait for a review from @davidc-gds before merging

davidc-gds:
Looks like sonar has flagged some unused imports that need removing. I wouldn't worry about the other issues.

app/src/main/kotlin/uk/govuk/app/ui/GovUkApp.kt
```diff
@@ -183,7 +188,9 @@ private fun BottomNavScaffold(
             color = GovUkTheme.colourScheme.surfaces.background
         ) {
             val startDestination =
-                if (shouldDisplayAnalyticsConsent) {
+                if (shouldDisplayUnavailable) {
```
davidc-gds, line 191: I've introduced an app launch navigation class to tidy this up. Might be worth waiting until my changes are merged and then integrating this with it.

app/build.gradle.kts
```diff
@@ -91,6 +91,7 @@ dependencies {
     implementation(projects.analytics)
     implementation(projects.config)
     implementation(projects.design)
+    implementation(projects.feature.availability)
```
davidc-gds, line 94: Not sure this needs to be a feature module, feels like more of a function of the app module itself?

TimGJD: Yes, I see what you mean. I've got 3 related screens going in (this unavailable one, forced update and recommend update) which is why I thought I'd pop them in a module together. I will move to main app module.

app/src/main/kotlin/uk/govuk/app/navigation/AppLaunchNavigation.kt
```diff
@@ -32,6 +32,10 @@ internal class AppLaunchNavigation(
         if (uiState.shouldDisplayAnalyticsConsent) {
             routeStack.push(ANALYTICS_GRAPH_ROUTE)
         }
+
+        if (uiState.shouldDisplayAppUnavailable) {
```
TimGJD, line 36: As App Unavailable has no further in-app navigation after it I have wondered if perhaps we shouldn't push the other graph routes in this scenario or maybe it doesn't matter and is simpler to leave as-is?

app/src/main/kotlin/uk/govuk/app/ui/GovUkApp.kt
TimGJD: Done

app/build.gradle.kts
TimGJD: Done

app/src/main/kotlin/uk/govuk/app/navigation/AppLaunchNavigation.kt
davidc-gds: I've had to re-jig this a bit anyway due to the bug we found during post amigo so I'll sort it out post merge

davidc-gds (approved):

PR #118, opened 2024-11-13
Visited items minor content fixups

gclssvglx:
Visited items: minor context fixes:

- Don't capitalise the word `month` in `This month`
- Remove `year` from dates
- Change day of month from 2 digits to 1 or 2, so `1 October` not `01 October`

### DIFF
--- a/feature/visited/src/main/kotlin/uk/govuk/app/visited/SectionTitles.kt
+++ b/feature/visited/src/main/kotlin/uk/govuk/app/visited/SectionTitles.kt
@@ -2,5 +2,5 @@ package uk.govuk.app.visited
 
 data class SectionTitles(
     val today: String = "Today",
-    val thisMonth: String = "This Month",
+    val thisMonth: String = "This month",
 )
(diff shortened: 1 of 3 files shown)


TimGJD (approved):

feature/visited/src/main/kotlin/uk/govuk/app/visited/SectionTitles.kt
```diff
@@ -2,5 +2,5 @@ package uk.govuk.app.visited
 
 data class SectionTitles(
     val today: String = "Today",
-    val thisMonth: String = "This Month",
+    val thisMonth: String = "This month",
```
TimGJD, line 5: Just a thought, do we need to consider having these as localisable?

gclssvglx: Ideally, yeah. They really should be res strings of course, but I just could not make that work as they are created as a hash `Today = [], This month = [], October 2023` etc before hitting the screen. If you know of a way I'm all ears, but we'd couldn't find a better solution at the time 🤷

PR #136, opened 2024-12-03
GOVUKAPP-993 - Capitalisation of months

gclssvglx:
# GOVUKAPP-993 - Capitalisation of months

The call to the `capitaliseMonth` function in the VisitedItemsTransformer somehow got removed at some point - add it back in.

## JIRA ticket(s)
  - [GOVAPP-993](https://govukverify.atlassian.net/browse/GOVUKAPP-993)

## Screen shots 

🎶  Note the use of fake data here!

| Before | After |
|---|---|
| month-before | month-after |

### DIFF
--- a/feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
+++ b/feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
@@ -21,7 +21,7 @@ fun transformVisitedItems(visitedItems: List<VisitedItemUi>, todaysDate: LocalDa
             it.toLocalDate().year != todaysDate.year
     }
 
-    fun VisitedItemUi.previousMonthKey() = "${toLocalDate().month.name} ${toLocalDate().year}"
+    fun VisitedItemUi.previousMonthKey() = "${capitaliseMonth(toLocalDate().month.name)} ${toLocalDate().year}"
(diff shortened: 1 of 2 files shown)


feature/visited/src/main/kotlin/uk/govuk/app/visited/data/VisitedItemsTransformer.kt
```diff
@@ -21,7 +21,7 @@ fun transformVisitedItems(visitedItems: List<VisitedItemUi>, todaysDate: LocalDa
             it.toLocalDate().year != todaysDate.year
     }
 
-    fun VisitedItemUi.previousMonthKey() = "${toLocalDate().month.name} ${toLocalDate().year}"
+    fun VisitedItemUi.previousMonthKey() = "${capitaliseMonth(toLocalDate().month.name)} ${toLocalDate().year}"
```
TimGJD, line 24: Just a thought that capitaliseMonth() (which I'm aware hasn't changed as part of this PR) is specifically named where it seems it's use could be for any string? It could be a string extension function with a more generic name for reuse with not just months maybe?

gclssvglx: Yeah, it could be generic you're right. I'm happy to rename - perhaps when we need it for something other than months.

PR #142, opened 2024-12-09
GOVUKAPP-1001: pages you've visited should ChildPageHeader

gclssvglx:
# GOVUKAPP-1001: pages you've visited should ChildPageHeader

ChildPageHeader component updated to handle optional Back (left hand side) button, 'action' (right hand side) button, and header text.

## JIRA ticket(s)
  - [GOVAPP-1001](https://govukverify.atlassian.net/browse/GOVUKAPP-1001)

## Figma
  - [Figma board](https://www.figma.com/design/5u5cvVCgKY3Rng2ADNV30K/2024-07---GOV.UK-app---beta-designs?node-id=1-83650&node-type=canvas&t=wNQf4guNUFZXVKUN-0)

## Screen shots 

| ChildPageHeader Previews |
|---|
| Screenshot 2024-12-09 at 11 53 21 |

| List Screen | Edit Screen |
|---|---|
| list-screen | edit-screen |

### DIFF
--- a/design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
+++ b/design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
@@ -42,20 +42,16 @@ fun TabHeader(
     }
 }
 
-data class ActionButton(
-    val text: String? = null,
-    val onClick: () -> Unit
-)
-
 @Composable
 fun ChildPageHeader(
     modifier: Modifier = Modifier,
     text: String? = null,
-    backButton: ActionButton? = null,
-    actionButton: ActionButton? = null
+    onBack: (() -> Unit)?,
+    onAction: (() -> Unit)?,
+    actionText: String? = null
 ) {
     Column(modifier) {
-        if (backButton != null || actionButton != null) {
+        if (onBack != null || onAction != null) {
             Row(
                 modifier = Modifier
                     .height(64.dp)
@@ -64,72 +60,51 @@ fun ChildPageHeader(
                 horizontalArrangement = Arrangement.SpaceBetween,
                 verticalAlignment = Alignment.CenterVertically,
             ) {
-                ChildPageHeaderBackButton(backButton)
-                ChildPageHeaderActionButton(actionButton)
+                if (onBack != null) {
+                    TextButton(
+                        onClick = onBack
+                    ) {
+                        Icon(
+                            imageVector = Icons.AutoMirrored.Filled.ArrowBack,
+                            contentDescription = stringResource(R.string.content_desc_back),
+                            tint = GovUkTheme.colourScheme.textAndIcons.link
+                        )
+                        Spacer(Modifier)
+                    }
+                }
+
+                if (onAction != null && actionText != null) {
+                    Spacer(Modifier)
+
+                    TextButton(
+                        onClick = onAction
+                    ) {
+                        BodyRegularLabel(
+                            text = actionText,
+                            color = GovUkTheme.colourScheme.textAndIcons.link,
+                            textAlign = TextAlign.End
+                        )
+                    }
+                }
             }
         }
 
         if (text != null) {
-            ChildPageHeaderText(text = text, modifier = modifier)
+            Row(
+                modifier = modifier
+                    .defaultMinSize(64.dp)
+                    .fillMaxWidth()
+                    .padding(GovUkTheme.spacing.small),
+                verticalAlignment = Alignment.CenterVertically,
+            ) {
+                LargeTitleBoldLabel(
+                    text = text,
+                    modifier = modifier
+                        .fillMaxWidth()
+                        .padding(horizontal = GovUkTheme.spacing.medium)
+                )
+            }
         }
-    }
-}
-
-@Composable
-private fun ChildPageHeaderBackButton(
-    actionButton: ActionButton? = null
-) {
-    if (actionButton == null) return
-
-    TextButton(
-        onClick = actionButton.onClick
-    ) {
-        Icon(
-            imageVector = Icons.AutoMirrored.Filled.ArrowBack,
-            contentDescription = stringResource(R.string.content_desc_back),
-            tint = GovUkTheme.colourScheme.textAndIcons.link
-        )
-        Spacer(Modifier)
-    }
-}
-
-@Composable
-private fun ChildPageHeaderActionButton(
-    actionButton: ActionButton? = null
-) {
-    if (actionButton == null) return
-
-    Spacer(Modifier)
-
-    TextButton(
-        onClick = actionButton.onClick
-    ) {
-        BodyRegularLabel(
-            text = actionButton.text!!,
-            color = GovUkTheme.colourScheme.textAndIcons.link,
-            textAlign = TextAlign.End
-        )
-    }
-}
-
-@Composable
-private fun ChildPageHeaderText(
-    text: String,
-    modifier: Modifier = Modifier
-) {
-    Row(
-        modifier = modifier
-            .defaultMinSize(64.dp)
-            .fillMaxWidth()
-            .padding(GovUkTheme.spacing.small),
-        verticalAlignment = Alignment.CenterVertically,
-    ) {
-        LargeTitleBoldLabel(
-            text = text,
-            modifier = modifier
-                .fillMaxWidth()
-                .padding(horizontal = GovUkTheme.spacing.medium)
-        )
     }
 }
 
@@ -146,8 +121,9 @@ private fun TabHeaderPreview() {
 private fun ChildPageHeaderNoTextWithBackAndActionPreview() {
     GovUkTheme {
         ChildPageHeader(
-            backButton = ActionButton(onClick = { }),
-            actionButton = ActionButton(text = "Done", onClick = { })
+            onBack = {},
+            onAction = {},
+            actionText = "Done"
         )
     }
 }
@@ -158,8 +134,9 @@ private fun ChildPageHeaderBackAndActionPreview() {
     GovUkTheme {
         ChildPageHeader(
             text = "Child page title",
-            backButton = ActionButton(onClick = { }),
-            actionButton = ActionButton(text = "Done", onClick = { })
+            onBack = { },
+            onAction = {},
+            actionText = "Done"
         )
     }
 }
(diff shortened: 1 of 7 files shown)


davidc-gds (dismissed):

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
```diff
@@ -40,27 +42,94 @@ fun TabHeader(
...
+                    .fillMaxWidth()
+                    .padding(GovUkTheme.spacing.small),
+                horizontalArrangement = Arrangement.SpaceBetween,
+                verticalAlignment = Alignment.CenterVertically,
             ) {
-                Icon(
-                    imageVector = Icons.AutoMirrored.Filled.ArrowBack,
-                    contentDescription = stringResource(R.string.content_desc_back),
-                    tint = GovUkTheme.colourScheme.textAndIcons.link
+                ChildPageHeaderBackButton(backButton, modifier = modifier.weight(1f))
```
TimGJD, line 67: These changes appear to cause back button touch area to cover width of screen if no action button is needed and touch area cover half width of screen if action button is needed. 
image
image

gclssvglx: Think that's the `Spacer`'s - I had issues aligning the icons without them. I'll take another look 👀

davidc-gds:
Can we resolve the Sonar issues relating to the modifiers and the preview function name (I think we're fine to leave the other issues)

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: I believe this is now resolved - thanks @TimGJD

gclssvglx:
Modifier issues resolved except:

- [Define a constant instead of duplicating this literal "Child page title" 4 times.](https://sonarcloud.io/project/issues?pullRequest=142&open=AZOrTkdOw4EDFRzKJaEE&id=alphagov_govuk-mobile-android-app)
- [This function has 8 parameters, which is greater than the 7 authorized.](https://sonarcloud.io/project/issues?pullRequest=142&open=AZOrTkb_w4EDFRzKJaEC&id=alphagov_govuk-mobile-android-app)

Is this ok @davidc-gds ?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
```diff
@@ -40,27 +42,94 @@ fun TabHeader(
...
+    val onClick: () -> Unit
+)
+
 @Composable
 fun ChildPageHeader(
-    text: String,
-    onBack: () -> Unit,
+    text: String? = null,
+    backButton: ActionButton? = null,
+    actionButton: ActionButton? = null,
```
davidc-gds: I'm not sure what this is giving us for the back button seeing as the text is not being used. Could we not just pass an onBack param and display the back button based on the presence of it?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
```diff
@@ -40,27 +42,94 @@ fun TabHeader(
...
+        }
+    }
+}
+
+@Composable
+private fun ChildPageHeaderBackButton(
+    actionButton: ActionButton? = null,
+    modifier: Modifier = Modifier
+) {
+    if (actionButton == null) return
```
davidc-gds, line 82: I'm not a fan of early returns from functions, what do you guys think?

TimGJD (approved):

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
```diff
@@ -40,27 +42,94 @@ fun TabHeader(
...
             }
         }
+
+        if (text != null) {
+            ChildPageHeaderText(text = text, modifier = modifier)
+        }
+    }
+}
+
+@Composable
```
davidc-gds, line 78: I'm torn on these private functions. Having all of the child page header functionality inside the one function seems nicer in these design library files but I get it makes sense to break things down into smaller chunks. Happy to discuss.

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: Happy to replace it if you know how to test for an empty lambda. I tried and searched, but from what I can see it looks like you can't 🤷  I came up with a data class because at least you can test for nullable on them.

I could create a BackAction data class - it'll only have the onClick of course?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: Personally, I like them - it's very Ruby idiomatic to do stuff like this. It also falls nicely into the fail-fast thinking - which I also personally like.

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: I felt  that the original version just had too much branching logic and `if` blocks - I guess I'm wired to break those out into functions. Happy to revert if that's the preference.

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
davidc-gds: You can make the lamda nullable:
onBack: (() -> Unit)?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
davidc-gds: It feels a bit weird to add the composable and then emit nothing from it. Rather just not adding it in the first place. It's a subjective thing though so you can leave it in.

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
davidc-gds: Ok, again, it's a subjective thing so you can leave it in. Maybe at some point we will have to extract the components into their own files.

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
TimGJD: For me it depends on the function (early return). In this case I would agree with @davidc-gds that it does not feel right to add a composable even if it's redundant?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: Refactored to use nullable lambda's (both onBack and onAction) - thanks @davidc-gds  👍

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: Refactored - see https://github.com/alphagov/govuk-mobile-android-app/pull/142#discussion_r1876308662

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: Refactored - see https://github.com/alphagov/govuk-mobile-android-app/pull/142#discussion_r1876308662

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
```diff
@@ -42,31 +44,67 @@ fun TabHeader(
 
 @Composable
 fun ChildPageHeader(
-    text: String,
-    onBack: () -> Unit,
     modifier: Modifier = Modifier,
+    text: String? = null,
+    onBack: (() -> Unit)?,
+    onAction: (() -> Unit)?,
```
TimGJD, line 50: We could default param to null here to save having to specify every time?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
```diff
@@ -42,31 +44,67 @@ fun TabHeader(
 
 @Composable
 fun ChildPageHeader(
-    text: String,
-    onBack: () -> Unit,
     modifier: Modifier = Modifier,
+    text: String? = null,
+    onBack: (() -> Unit)?,
```
TimGJD, line 49: We could default param to null here to save having to specify every time?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: Great suggestion - thanks @TimGJD 👍 - fixed up

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
gclssvglx: Great suggestion - thanks @TimGJD 👍 - fixed up

feature/visited/src/main/kotlin/uk/govuk/app/visited/ui/VisitedScreen.kt
```diff
@@ -86,55 +80,24 @@ private fun VisitedScreen(
...
-                TextButton(
-                    onClick = onBack,
-                ) {
-                    Icon(
-                        imageVector = Icons.AutoMirrored.Filled.ArrowBack,
-                        contentDescription = stringResource(uk.govuk.app.design.R.string.content_desc_back),
-                        tint = GovUkTheme.colourScheme.textAndIcons.link
-                    )
-                }
+            var onAction: (() -> Unit)? = null
```
TimGJD, line 83: We could make this a val instead? 

```
val onAction = if (!visitedItems.isNullOrEmpty()) {
                onEditClick
            } else {
                null
            }
```

TimGJD (dismissed):

davidc-gds (approved):

davidc-gds (approved):

PR #149, opened 2024-12-11
GOVUKAPP-994 Talkback reads out wrong item

TimGJD:
# Talkback reads out wrong item

When selecting an item on PYV screen talkback reads out the wrong item

# Resolution
- Add delay to reordering of PYV items which is cause of list item being re-read after reordering
- Set focus on PYV screen title every time app lifecycle calls resume (on initial screen landing and on returning to the app as requested by design)

## JIRA ticket(s)
  - [GOVAPP-994](https://govukverify.atlassian.net/browse/GOVUKAPP-994)

## Figma
  - [Figma board](https://www.figma.com/design/5u5cvVCgKY3Rng2ADNV30K/2024-07---GOV.UK-app---beta-designs?node-id=1-83658&node-type=section&t=GobhzpuPdcZLiIEH-0)

## Video 
https://github.com/user-attachments/assets/48aa46b2-d621-4bd6-8377-6f6727219d0b


feature/visited/src/main/kotlin/uk/govuk/app/visited/VisitedViewModel.kt
```diff
@@ -123,6 +124,7 @@ internal class VisitedViewModel @Inject constructor(
 
     fun onVisitedItemClicked(title: String, url: String) {
         viewModelScope.launch {
+            delay(1000)
```
TimGJD, line 127: It has crossed my mind we could check and only add this if the user has accessibility enabled?

gclssvglx (dismissed):

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
```diff
@@ -46,7 +49,8 @@ fun ChildPageHeader(
     text: String? = null,
     onBack: (() -> Unit)? = null,
     onAction: (() -> Unit)? = null,
-    actionText: String? = null
+    actionText: String? = null,
+    focusRequester: FocusRequester? = null
```
davidc-gds, line 53: Can we not do this via the modifier we pass in i.e. request focus for the whole header (that way this very specific code would live where it is most relevant)? Rather than passing in the focus requester and applying it to only the title?

feature/visited/src/main/kotlin/uk/govuk/app/visited/VisitedViewModel.kt
davidc-gds: I think it would make more sense for this delay to happen in the composable rather than the view model as the issue it UI related?

design/src/main/kotlin/uk/govuk/app/design/ui/component/Header.kt
TimGJD: @davidc-gds yeah, that's what I did initially (made the ChildPageHeader focusable and set focus on it) but it didn't take focus. I also tried merging descendants but then the whole ChildPageHeader is focussed. This was only solution I could find.

davidc-gds: TBH in that case I think it would make more sense to use a custom label for the title on the visited screen rather than add that functionality to the child page header. We could still use the child page header for the back button and action button, just not give it a title. Still not perfect but similar approach to topics and subtopics (not scrollable though). What do you think @TimGJD @gclssvglx ?

gclssvglx: @davidc-gds personally, I think the TalkBack issue is super edge-case and can be mostly resolved by using a delay. Using components on the other hand has got to be the preferred long-term approach and is more sustainable than edge-case fixes for what are essentially TalkBack bugs.  So, on balance my preference would be to keep the ChildPageHeader and add just the delay. Log it as a known issue and hope that it's resolved where is should be - in TalkBack.

My concern is that if we keep putting patch after patch in, the code will start to get unmaintainable.

gclssvglx (dismissed):

davidc-gds (approved):

davidc-gds (approved):

PR #179, opened 2025-01-22
GOVUKAPP-1007: Reduce api call duplication

davidc-gds:
# Title

Reduce api call duplication

## JIRA ticket(s)
  - [GOVUKAPP-1007](https://govukverify.atlassian.net/browse/GOVUKAPP-1007)


TimGJD (commented):
Can the file ConfigExceptions.kt be deleted with these changes?

design/src/main/kotlin/uk/govuk/app/design/ui/component/error/ErrorConstants.kt
```diff
@@ -0,0 +1,5 @@
+package uk.govuk.app.design.ui.component.error
+
+object ErrorConstants {
+    const val GOV_UK_URL = "https://www.gov.uk"
```
gclssvglx, line 4: This is all over the place - in different modules - can we centralise somehow?

app/src/main/kotlin/uk/govuk/app/ui/AppUnavailableScreen.kt
```diff
@@ -16,8 +16,9 @@ import uk.govuk.app.design.ui.component.LargeTitleBoldLabel
 import uk.govuk.app.design.ui.component.ListDivider
 import uk.govuk.app.design.ui.component.MediumVerticalSpacer
 import uk.govuk.app.design.ui.component.SecondaryButton
+// Todo - this feels weird - we need a better way to share constants between modules
```
gclssvglx: Ah! Just noticed this - yes please!

design/src/main/kotlin/uk/govuk/app/design/ui/component/error/ErrorConstants.kt
davidc-gds: I think I have as it is now only used in the design module. Not sure it would be wise to link this to the base urls of search, privacy police, terms etc.

gclssvglx: Ok - yeah linking the others is what I was thinking of. Sounds like it's not a good idea! Happy to revisit later.

gclssvglx (approved):
Looks good 👍

design/src/main/kotlin/uk/govuk/app/design/ui/component/error/ErrorConstants.kt
davidc-gds: I'm just thinking that it's more of a coincidence that they are the same more than anything else and therefore not unlikely to change (especially search)?

TimGJD (approved):

PR #186, opened 2025-01-29
Maestro is unable to find and interact with our ToggleSwitch component implementation

gclssvglx:
# Maestro is unable to find and interact with our ToggleSwitch component implementation

This effectively means that we are unable to create Maestro tests that either assert on a particular state , or seek to update the current state of these elements.

This change adds a `resource-id` which uniquely identifies each ToggleSwitch instance.

## JIRA ticket(s)
  - [GOVAPP-1107](https://govukverify.atlassian.net/browse/GOVUKAPP-1107)

### DIFF
--- a/design/src/main/kotlin/uk/govuk/app/design/ui/component/Switch.kt
+++ b/design/src/main/kotlin/uk/govuk/app/design/ui/component/Switch.kt
@@ -3,14 +3,20 @@ package uk.govuk.app.design.ui.component
 import androidx.compose.material3.Switch
 import androidx.compose.material3.SwitchDefaults
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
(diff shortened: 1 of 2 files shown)


design/src/main/kotlin/uk/govuk/app/design/ui/component/Switch.kt
```diff
@@ -3,14 +3,20 @@ package uk.govuk.app.design.ui.component
...
+import androidx.compose.ui.semantics.testTagsAsResourceId
 import androidx.compose.ui.tooling.preview.Preview
 import uk.govuk.app.design.ui.theme.GovUkTheme
 
+@OptIn(ExperimentalComposeUiApi::class)
 @Composable
 fun ToggleSwitch(
     checked: Boolean,
     onCheckedChange: (Boolean) -> Unit,
+    testDescription: String,
```
TimGJD, line 19: Another option could be just to pass the configured modifier in the ToggleSwitch constructor but I guess either way is fine? I think both ways have different benefits.

TimGJD (approved):

PR #187, opened 2025-01-29
GOVUKAPP-1114: Step by steps API call

davidc-gds:
# Title

Cache step by steps when fetching topic

## JIRA ticket(s)
  - [GOVUKAPP-1114](https://govukverify.atlassian.net/browse/GOVUKAPP-1114)

[GOVUKAPP-1114]: https://govukverify.atlassian.net/browse/GOVUKAPP-1114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature/topics/src/main/kotlin/uk/govuk/app/topics/data/TopicsRepo.kt
+++ b/feature/topics/src/main/kotlin/uk/govuk/app/topics/data/TopicsRepo.kt
@@ -18,7 +18,10 @@ internal class TopicsRepo @Inject constructor(
     private val localDataSource: TopicsLocalDataSource,
 ) {
 
-    private lateinit var stepBySteps: List<RemoteTopicContent>
+    private lateinit var _stepBySteps: List<RemoteTopicContent>
+
+    val stepBySteps
+        get() = _stepBySteps
 
     suspend fun sync(): Boolean {
         var success = false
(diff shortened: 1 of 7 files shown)


feature/topics/src/main/kotlin/uk/govuk/app/topics/data/TopicsRepo.kt
```diff
@@ -17,6 +18,11 @@ internal class TopicsRepo @Inject constructor(
     private val localDataSource: TopicsLocalDataSource,
 ) {
 
+    private lateinit var _stepBySteps: List<RemoteTopicContent>
+
+    val stepBySteps
+        get() = _stepBySteps
```
TimGJD, line 24: Could this be open to an exception?

davidc-gds: Yes but IMO that would be preferable to making it nullable and having to handle the null. The state in which it would be null or uninitialised would be a dev error i.e. trying to access it before fetching a topic so my thinking is better to fail fast and flag the issue rather than fail silently.

TimGJD (approved):

PR #223, opened 2025-03-24
GOVUKAPP-1348 : local fake api integration

gclssvglx:
# GOVUKAPP-1348 : local fake api integration

## 🎵  Notes for reviewers

Read the [development API documentation](https://github.com/gclssvglx/development-local-api/blob/main/README.md) first. It explains the structure of the API and JSON responses. These will be exactly the same as those delivered by the production API.

Once you've got the app running here are some postcodes to try...

- `E1 8QS` - gives a Unitary (single Local Authority) response
- `DE45 1QW` - gives a Country/District (two-tier Local Authority) response
- `BH22 8UB` - gives an ambiguous response. Here a list of addresses is presented as the list for that postcode includes 2 or more distinct Local Authorities.
- All invalid postcodes (e.g. `SW1A`) return a 400 `Invalid postcode` response
- All valid, but not found in the development API data (e.g. `SW1A 1AA`) return a 404 `Not found` response

You will notice that 400 and 404 statuses are returned for invalid and not found postcode responses. This is due to the desire to return semantically correct HTTP statuses. 

Also you may notice that the postcode API call redirects to the slug API call when possible.  

The `For you` header on the local homepage widget is missing. This will be added (if still required) later but is omitted at this stage in order to keep the development of this feature on-time.

## JIRA ticket(s)
  - [GOVUKAPP-GOVUKAPP-1348](https://govukverify.atlassian.net/browse/GOVUKAPP-1348)

## Figma
  - [Figma board](https://www.figma.com/design/N4v2GPQhwZ7DwhqucyJqlu/GOV.UK-apps-and-notifications?node-id=20393-21993&p=f&t=l0RASoIsOHB7NGKG-0)


feature/local/src/main/kotlin/uk/govuk/app/local/data/remote/LocalApiCall.kt
davidc-gds, line 1: Is there any reason that we can't use the existing safe api call we have?

design/src/main/kotlin/uk/gov/govuk/design/ui/theme/Color.kt
davidc-gds, line 1: Have these colours been added as tokens in Figma?

feature/local/src/main/kotlin/uk/govuk/app/local/data/remote/model/ApiResponseAdapter.kt
davidc-gds, line 1: Why do we need to perform custom JSON parsing?

feature/local/src/main/kotlin/uk/govuk/app/local/data/remote/LocalApiCall.kt
gclssvglx: Yeah, I tried that - but adding the ApiResponse.MessageResponse data class creates a cyclic-dependency. I could add a generic Response class - but that's have to live in the data module. Would that be a better approach?

design/src/main/kotlin/uk/gov/govuk/design/ui/theme/Color.kt
gclssvglx: I'll check with the design team.

gclssvglx: We believe the `GreenLighter50` colour is there. 

These will all be added tomorrow. If the names don't match the convention - I'll update them in the code.

text and icons tokens: 
  - localIcon

surfaces tokens: 
  - localBackground
  - textFieldBackground
  - highlightedTextField

strokes tokens:
  - localBorder
  - textFieldCursor

feature/local/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <uses-permission android:name="android.permission.INTERNET" />
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+
+  <!-- TODO: REMOVE THIS WHEN THE PRODUCTION LOCAL API IS IN AVAILABLE -->
+  <application
+    android:usesCleartextTraffic="true">
```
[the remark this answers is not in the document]
gclssvglx: Interesting, I didn't know that was possible thanks. Pragmatically, this won't be here too long - from there on we can use integration for development, with is SSL. So, if you're ok - I'll leave this here until the integration instance is stood up.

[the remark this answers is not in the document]
gclssvglx: This has now been removed as it's no longer needed!

feature/local/src/main/kotlin/uk/govuk/app/local/data/remote/model/ApiResponseAdapter.kt
gclssvglx: The need for custom JSON parsing has been removed.

feature/local/src/main/kotlin/uk/govuk/app/local/data/remote/LocalApiCall.kt
gclssvglx: As per your suggestion - the ApiResponse is now much simpler...

```kotlin
data class ApiResponse(
    @SerializedName("local_authority") val localAuthority: LocalAuthority?,
    @SerializedName("addresses") val addresses: List<Address>?,
    @SerializedName("message") val message: String?
)
```

But, we still need the custom safe API call function to handle the 400 and 404's.

davidc-gds (approved):

PR #231, opened 2025-04-09
GOVUKAPP-1356: Store local authority

gclssvglx:
# GOVUKAPP-1356: Store local authority

## JIRA ticket(s)
  - [GOVUKAPP-1356](https://govukverify.atlassian.net/browse/GOVUKAPP-1356)

[GOVUKAPP-1356]: https://govukverify.atlassian.net/browse/GOVUKAPP-1356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
+++ b/feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
@@ -31,8 +31,8 @@ internal class LocalRepo @Inject constructor(
     }
 
     private suspend fun processResponse(response: Result<ApiResponse>) {
-        if (response is Result.Success && response.value.localAuthority != null) {
-            val localAuthority = response.value.localAuthority
+        if (response is Result.Success) {
+            val localAuthority = response.value.localAuthority ?: return
 
             storeLocalAuthorities(
                 localAuthorities = arrayListOf(
--- a/feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
+++ b/feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
@@ -32,27 +32,23 @@ internal class LocalRepo @Inject constructor(
 
     private suspend fun processResponse(response: Result<ApiResponse>) {
         if (response is Result.Success) {
-            val localAuthority = response.value.localAuthority ?: return
-
-            storeLocalAuthorities(
-                localAuthorities = arrayListOf(
-                    localAuthority,
-                    localAuthority?.parent
+            response.value.localAuthority?.let { localAuthority ->
+                storeLocalAuthorities(
+                    child = localAuthority,
+                    parent = localAuthority.parent
                 )
-            )
+            }
         }
     }
 
-    private suspend fun storeLocalAuthorities(localAuthorities: List<LocalAuthority?>) {
-        val localAuthority = localAuthorities[0] ?: return
-        val localAuthorityParent = localAuthorities[1]
+    private suspend fun storeLocalAuthorities(child: LocalAuthority, parent: LocalAuthority?) {
         localDataSource.insertOrReplace(
-            localAuthority.name,
-            localAuthority.homePageUrl,
-            localAuthority.slug,
-            localAuthorityParent?.name,
-            localAuthorityParent?.homePageUrl,
-            localAuthorityParent?.slug
+            child.name,
+            child.homePageUrl,
+            child.slug,
+            parent?.name,
+            parent?.homePageUrl,
+            parent?.slug
         )
     }
 }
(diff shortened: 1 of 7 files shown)


feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
```diff
@@ -3,23 +3,61 @@ package uk.govuk.app.local.data
...
+        val response = safeLocalApiCall { localApi.getLocalAuthority(slug) }
+        processResponse(response)
+        return response
+    }
+
+    private suspend fun processResponse(response: Result<ApiResponse>) {
+        if (response is Result.Success && response.value.localAuthority != null) {
+            val localAuthority = response.value.localAuthority
+
+            var parentLocalAuthority: LocalAuthority? = null
```
TimGJD, line 37: It looks like this var and logic to assign it could be unnecessary and we could just use `localAuthority?.parent` instead below?

feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
```diff
@@ -3,23 +3,61 @@ package uk.govuk.app.local.data
...
+                localAuthorities = arrayListOf(
+                    localAuthority,
+                    parentLocalAuthority
+                )
+            )
+        }
+    }
+
+    private suspend fun storeLocalAuthorities(localAuthorities: List<LocalAuthority?>) {
+        if (localAuthorities.isNotEmpty()) {
```
TimGJD, line 52: Not empty check would cover element 0 from an index out of bounds exception but not element 1?

feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
```diff
@@ -3,23 +3,61 @@ package uk.govuk.app.local.data
...
+            localDataSource.insertOrReplace(
+                localAuthorities[0]?.name ?: "",
```
TimGJD, line 54: Suggest the below might be cleaner to reduce all the null checks?
```
val local = localAuthorities[0] ?: return
val localParent = localAuthorities[1]
            localDataSource.insertOrReplace(
                local.name,
                local.homePageUrl,
                local.slug,
                localParent?.name,
                localParent?.homePageUrl,
                localParent?.slug
            )
```

feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
```diff
@@ -3,23 +3,61 @@ package uk.govuk.app.local.data
...
     ): Result<ApiResponse> {
-        return safeLocalApiCall { localApi.getLocalAuthority(slug) }
+        val response = safeLocalApiCall { localApi.getLocalAuthority(slug) }
+        processResponse(response)
+        return response
+    }
+
+    private suspend fun processResponse(response: Result<ApiResponse>) {
+        if (response is Result.Success && response.value.localAuthority != null) {
+            val localAuthority = response.value.localAuthority
```
TimGJD, line 35: We could remove null check out of `if` statement and just do `val localAuthority = response.value.localAuthority ?: return` here instead?

gclssvglx: Fixed up. Thanks

feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
```diff
@@ -3,23 +3,56 @@ package uk.govuk.app.local.data
...
+    }
+
+    private suspend fun processResponse(response: Result<ApiResponse>) {
+        if (response is Result.Success) {
+            val localAuthority = response.value.localAuthority ?: return
+
+            storeLocalAuthorities(
+                localAuthorities = arrayListOf(
+                    localAuthority,
+                    localAuthority?.parent
```
TimGJD, line 40: The new null check on line 35 should make this new null check here redundant now?

feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
```diff
@@ -3,23 +3,56 @@ package uk.govuk.app.local.data
...
+                )
+            )
+        }
+    }
+
+    private suspend fun storeLocalAuthorities(localAuthorities: List<LocalAuthority?>) {
+        val localAuthority = localAuthorities[0] ?: return
```
TimGJD, line 47: For IOB best practice add size check on localAuthorities back in here to cover 2 elements?

davidc-gds: Fixed

feature/local/src/main/kotlin/uk/govuk/app/local/data/LocalRepo.kt
davidc-gds: Fixed

PR #237, opened 2025-04-16
GOVUKAPP-1352: Local authority widgets

davidc-gds:
# Title

Description

## JIRA ticket(s)
  - [GOVUKAPP-1352](https://govukverify.atlassian.net/browse/GOVUKAPP-1352)

## Figma
  - [Figma board](https://www.figma.com/design/N4v2GPQhwZ7DwhqucyJqlu/GOV.UK-apps-and-notifications?node-id=20393-21993&p=f&t=qeAkecccbBPloZYI-0)

## Screen shots

Screenshot_20250416_102255
Screenshot_20250416_102336
Screenshot_20250416_102704

[GOVUKAPP-1352]: https://govukverify.atlassian.net/browse/GOVUKAPP-1352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


design/src/main/kotlin/uk/gov/govuk/design/ui/component/Card.kt
```diff
@@ -136,6 +139,53 @@ fun HomeNavigationCard(
     }
 }
 
+@Composable
+fun SearchResultCard(
```
gclssvglx, line 143: I wonder if `SearchResultCard` is the best name for this 🤔  Would something like `ClickableLinkCard` be better, given it's used by Search and now Local?

gclssvglx (approved):
Fantastic work 🥇 

One small (ignorable) suggestion around the new card name - not a blocker

design/src/main/kotlin/uk/gov/govuk/design/ui/component/Card.kt
davidc-gds: I thought exactly the same thing but that's what it's called in the design lib. I'll raise it with Mike and see if they're happy to update the name to minimise confustion.

PR #238, opened 2025-04-16
GOVUKAPP-1135 Catch exceptions from Gson

TimGJD:
# Catch exceptions from Gson

- Add exception handling for Gson when parsing JSON to Additional Data.

## JIRA ticket(s)
  - [GOVUKAPP-1135](https://govukverify.atlassian.net/browse/GOVUKAPP-1135)

[GOVUKAPP-1135]: https://govukverify.atlassian.net/browse/GOVUKAPP-1135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


notifications/src/main/kotlin/uk/gov/govuk/notifications/model/AdditionalData.kt
```diff
@@ -2,10 +2,23 @@ package uk.gov.govuk.notifications.model
...
+ *
+ * If successful returns an Additional Data object.
+ * If unsuccessful returns null.
+ */
+internal fun JSONObject?.asAdditionalData(): AdditionalData? {
+    return try {
+        val gson = Gson()
+        gson.fromJson(this.toString(), AdditionalData::class.java)
+    } catch (_: Exception) {
+        null
```
davidc-gds, line 22: Do we need any form of indication if this is happening a lot in order to flag issues with the creation of the notifications?

TimGJD: Both iOS and Android fail silently if they cannot parse the additional data. You've raised a good point which I am going to raise separate iOS and Android tickets for and take to design so they can decide the behaviour if they want Page Not Found or something else.

davidc-gds (approved):

PR #261, opened 2025-05-08
Upload latest to Google Play version name 0.1.0

TimGJD:
# Upload latest to Google Play version name 0.1.0


davidc-gds:
I think we should rename the release branch to 0.1?

PR #287, opened 2025-05-21
GOVUKAPP-1604 Notifications permission navigation

TimGJD:
# Notifications permission navigation

- Add finish function to view models to avoid callbacks being fired twice
- Add remove consent function and check and set it on app launch and on notification permission changes

## JIRA ticket(s)
  - [GOVUKAPP-1604](https://govukverify.atlassian.net/browse/GOVUKAPP-1604)

[GOVUKAPP-1604]: https://govukverify.atlassian.net/browse/GOVUKAPP-1604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/notifications/src/main/kotlin/uk/gov/govuk/notifications/NotificationsConsentViewModel.kt
+++ b/notifications/src/main/kotlin/uk/gov/govuk/notifications/NotificationsConsentViewModel.kt
@@ -24,7 +24,10 @@ internal class NotificationsConsentViewModel @Inject constructor(
         status: PermissionStatus
     ) {
         viewModelScope.launch {
-            _uiState.value = if (!status.isGranted || notificationsClient.consentGiven()) {
+            _uiState.value = if (!status.isGranted) {
+                notificationsClient.removeConsent()
+                NotificationsUiState.Finish
+            } else if (notificationsClient.consentGiven()) {
                 NotificationsUiState.Finish
             } else {
                 NotificationsUiState.Default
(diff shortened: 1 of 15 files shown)


notifications/src/main/kotlin/uk/gov/govuk/notifications/NotificationsConsentViewModel.kt
```diff
@@ -31,4 +31,8 @@ internal class NotificationsConsentViewModel @Inject constructor(
             }
         }
     }
+
+    internal fun finish() {
+        _uiState.value = NotificationsUiState.Finish
+    }
```
gclssvglx, line 40: I know if there was a better way you would have done that. But this does seem a little crude.

gclssvglx (approved):

notifications/src/main/kotlin/uk/gov/govuk/notifications/NotificationsConsentViewModel.kt
TimGJD: Yeah, the current notifications requirements have been tricky, particularly coming up with a solution to cover pre and post Android 13. I had to come up with a solution to avoid code duplication as sonar blocked a previous PR and this is the safest way I could find so far that appears to be bug free.

PR #296, opened 2025-06-02
GOVUKAPP-1560: Attestation

davidc-gds:
# Title

Fetch attestation token from Firebase app check and include as a header in auth and refresh.

## JIRA ticket(s)
  - [GOVUKAPP-1560](https://govukverify.atlassian.net/browse/GOVUKAPP-1560)

[GOVUKAPP-1560]: https://govukverify.atlassian.net/browse/GOVUKAPP-1560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


data/src/main/kotlin/uk/gov/govuk/data/auth/AuthRepo.kt
```diff
@@ -145,31 +177,19 @@ class AuthRepo @Inject constructor(
...
-        tokenResponse: TokenResponse?,
-        exception: AuthorizationException?,
-        refreshToken: String? = null
-    ): Boolean {
-        val mappedTokenResponse = tokenResponseMapper.map(tokenResponse)
-        val accessToken = mappedTokenResponse.accessToken
-        val idToken = mappedTokenResponse.idToken
-        val mappedRefreshToken = refreshToken ?: mappedTokenResponse.refreshToken
+    fun isAuthenticationEnabled(): Boolean {
+        val result  = biometricManager.canAuthenticate(
```
TimGJD, line 180: Formatting here?

data/src/main/kotlin/uk/gov/govuk/data/auth/AuthRepo.kt
```diff
@@ -77,54 +74,89 @@ class AuthRepo @Inject constructor(
...
     }
 
-    suspend fun refreshTokens(
+    suspend fun persistRefreshToken(
         activity: FragmentActivity,
         title: String,
         subtitle: String? = null,
         description: String? = null
     ): Boolean {
+        secureStore.upsert(REFRESH_TOKEN_KEY, tokens.refreshToken)
```
TimGJD, line 146: Do we need to catch exceptions from upsert? https://github.com/govuk-one-login/mobile-android-secure-store/blob/833aa16fa412ef4e6bca20b4aae117277a62be02/app/src/main/java/uk/gov/android/securestore/SecureStore.kt#L44

data/src/main/kotlin/uk/gov/govuk/data/auth/AuthRepo.kt
davidc-gds: Formatting looks fine in the actual file, maybe it looks weird in the diff?

data/src/main/kotlin/uk/gov/govuk/data/auth/AuthRepo.kt
davidc-gds: We probably should for all secure store functions that throw. I don't think that is in the scope of this PR though as that is all existing code (all I've done is move the functions around in the class to make it more readable). Maybe we should make a note and raise as tech debt? Good spot!

TimGJD (approved):

gclssvglx (approved):

PR #304, opened 2025-06-09
Chat POC

gclssvglx:
## JIRA ticket(s)
  - [GOVUKAPP-1840](https://govukverify.atlassian.net/browse/GOVUKAPP-1840)
  - [GOVUKAPP-1849](https://govukverify.atlassian.net/browse/GOVUKAPP-1849)


feature/chat/src/main/kotlin/uk/gov/govuk/chat/data/ChatRepo.kt
```diff
@@ -0,0 +1,117 @@
...
+            userQuestion = question
+        )
+
+        try {
+            // "200": Success
+            val response = chatApi.startConversation(requestBody)
+            _conversationId = response.conversationId
+            println(response)
+        } catch (e: Exception) {
+            // TODO: handle errors
```
harrytmthy-dev, line 31: Hi @gclssvglx 👋 Just a thought on the repeated try-catch blocks for API calls: you could reduce duplication and centralize error handling using a simple wrapper like:

```kotlin
private suspend fun <T> safeApiCall(action: suspend () -> T): T? {
    return try {
        withContext(Dispatchers.IO) { action() }
    } catch (e: Exception) {
        // TODO: handle errors
        // "422": Validation error on question submission (such as PII in question)
        // "429": Too many requests to read endpoints from an Api User or Device ID
        println("API error: ${e.message}")
        null
    }
}
```

Then in your repo functions:

```kotlin
val response = safeApiCall { chatApi.startConversation(requestBody) }
```

Could help reduce boilerplate and keep things consistent, especially if more endpoints or error cases get added later on. Let me know what you think 🙌

davidc-gds (approved):

PR #374, opened 2025-07-23
GOVUKAPP-2041 : Post Amigo fixups

gclssvglx:
Fixes the following issues:

- When PII is detected the submit button should be disabled
- Add alt text for action menu button: 'More options'
- Add alt text for 'submit' button: 'Send message'
- Change focus order so that the text input box is focused on before the action menu button for Talkback

## JIRA ticket(s)
  - [GOVUKAPP-2041](https://govukverify.atlassian.net/browse/GOVUKAPP-2041)

[GOVUKAPP-2041]: https://govukverify.atlassian.net/browse/GOVUKAPP-2041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/ChatScreen.kt
+++ b/feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/ChatScreen.kt
@@ -239,7 +239,7 @@ private fun ChatContent(
 
                             SubmitIconButton(
                                 onClick = { onSubmit(uiState.question) },
-                                enabled = uiState.isSubmitEnabled
+                                uiState = uiState
                             )
                         }
                     }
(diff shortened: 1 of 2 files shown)


feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/ChatScreen.kt
```diff
@@ -148,6 +151,7 @@ private fun ChatContent(
                     modifier = Modifier
                         .background(GovUkTheme.colourScheme.surfaces.chatBackground)
                         .padding(all = GovUkTheme.spacing.medium)
+                        .semantics { isTraversalGroup = true }
```
gclssvglx, line 154: From everything I've seen this implementation should be correct, but it does not work as desired. I've tried multiple approaches to rectify: new AVD, clear caches, different AVD, different API level, wipe AVD data, removing the AnimatedVisibility, removing everything but the action button and the text input ... Nothing seems to solve it. I've created a minimal test - that works, so It must be something in the code - I just can't seem to find it.

TimGJD (dismissed):

PR #455, opened 2025-10-01
GOVUKAPP-2605 : Change polling frequency to support decimal

gclssvglx:
Change data type of `chatPollIntervalSeconds` from `Int` to `Double` in order to support sub-second intervals.

## JIRA ticket(s)
  - [GOVUKAPP-2605](https://govukverify.atlassian.net/browse/GOVUKAPP-2605)

[GOVUKAPP-2605]: https://govukverify.atlassian.net/browse/GOVUKAPP-2605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/src/main/kotlin/uk/gov/govuk/config/data/ConfigRepo.kt
+++ b/config/src/main/kotlin/uk/gov/govuk/config/data/ConfigRepo.kt
@@ -21,7 +21,7 @@ class ConfigRepo @Inject constructor(
 ) {
 
     companion object {
-        private const val CHAT_POLL_INTERVAL_FALLBACK = 3
+        private const val CHAT_POLL_INTERVAL_FALLBACK = 3.0
     }
 
     private lateinit var _config: Config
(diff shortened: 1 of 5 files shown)


config/src/main/kotlin/uk/gov/govuk/config/data/ConfigRepo.kt
```diff
@@ -60,6 +60,6 @@ class ConfigRepo @Inject constructor(
         }
     }
 
-    val chatPollIntervalSeconds: Int
-        get() = config.chatPollIntervalSeconds?.takeIf { it > 0 } ?: CHAT_POLL_INTERVAL_FALLBACK
+    val chatPollIntervalSeconds: Double
```
TimGJD, line 63: Would it make more sense for this to be a Float as it looks like we don't need the precision of a Double here?

gclssvglx: I'd prefer to stick with Double - at least for now

PR #676, opened 2026-04-13
GOVUKAPP-3400 Update launch mode

TimGJD:
# Update launch mode

- Update the app launch mode and remove intent flags which we added for Compose deep linking which we now no longer use to resolve Custom Chrome Tabs launching as a separate task when they should be launching within the context of the app (see before/after screen shots below).

- Drop the oldest intent to resolve deep link data not being read when the app is opened from a deep link whilst running in the background. OnNewIntent is called twice in this scenario, the first time with no intent data and the second time has intent data so dropping the oldest intent means new subscribers will get the latest intent (with data) when added. Thanks to @lamprosgk for this suggestion.

_I have tested this on emulator on Android 16 and physical Android 10 device but would appreciate others testing also if you have the time._

## JIRA ticket(s)
  - [GOVUKAPP-3400](https://govukverify.atlassian.net/browse/GOVUKAPP-3400)

## Screen shots

| Before | After |
|--------|-------|
| image | image |

### DIFF
--- a/app/src/main/kotlin/uk/gov/govuk/MainActivity.kt
+++ b/app/src/main/kotlin/uk/gov/govuk/MainActivity.kt
@@ -1,8 +1,6 @@
 package uk.gov.govuk
 
 import android.content.Intent
-import android.content.Intent.FLAG_ACTIVITY_CLEAR_TASK
-import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import android.os.Bundle
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/uk/gov/govuk/MainActivity.kt
```diff
@@ -28,7 +26,7 @@ class MainActivity : FragmentActivity() {
     internal lateinit var appNavigation: AppNavigation
 
     private val _intentFlow: MutableSharedFlow<Intent> =
-        MutableSharedFlow(replay = 1)
+        MutableSharedFlow(replay = 2)
```
lamprosgk: I think it's better to keep `replay = 1` and add overwrite the oldest emission flag with `onBufferOverflow = BufferOverflow.DROP_OLDEST`, subscribers would get single, most recent intent

TimGJD: Ah yes, that is much better! Thanks

lamprosgk (approved):

PR #681, opened 2026-04-17
GOVUKAPP-3394 Addressing post amigos feedback for account connection card

lamprosgk:
# Title

Addressing post amigos feedback 
https://gds.slack.com/archives/C0851ER905P/p1776246345197379

## JIRA ticket(s)
  - [GOVUKAPP-3394](https://govukverify.atlassian.net/browse/GOVUKAPP-3394)

Screenshot_20260417_140011

[Screen_recording_20260417_140937.webm](https://github.com/user-attachments/assets/f05ba1fc-aacb-4b30-b5d5-6326b7919ea4)


design/src/main/kotlin/uk/gov/govuk/design/ui/component/Label.kt
```diff
@@ -62,15 +63,27 @@ private fun BaseLabel(
...
-    )
+    val altText = text.text
+        .replace(
+            stringResource(R.string.gov_uk),
+            stringResource(R.string.gov_uk_alt_text)
+        )
+        // replace acronyms with alt text so Talkback announces properly - for example MOT as M.O.T. and not 'moat'
+        .replaceAcronym(
+            stringResource(R.string.acronym_mot),
+            stringResource(R.string.acronym_mot_alt_text)
```
TimGJD: This feels very specific to driving to be in design module and generic ui component? 
Could we end up with loads of these in here eventually for different acronyms?
Wondering if we should have an optional altText property for this composable instead?

lamprosgk: Yeh you're right, good point 👍🏼

TimGJD (approved):

PR #683, opened 2026-04-20
GOVUKAPP-3220: Realm migration

davidc-gds:
# Title

Migrate from Realm to Room

## JIRA ticket(s)
  - [GOVUKAPP-3220](https://govukverify.atlassian.net/browse/GOVUKAPP-3220)


data/src/main/kotlin/uk/gov/govuk/data/local/RoomEncryptionHelper.kt
```diff
@@ -0,0 +1,50 @@
...
+class RoomEncryptionHelper @Inject constructor(
+    @ApplicationContext private val context: Context
+) {
+
+    private companion object {
+        private const val PREFS_NAME = "room_keyset_prefs"
+        private const val KEYSET_NAME = "room_keyset"
+        private const val MASTER_KEY_URI = "android-keystore://room_master_key"
+        private const val DB_KEY_PREF = "room_db_key"
+        private const val KEY_SIZE_BYTES = 32
```
gclssvglx, line 19: I wonder if this is a little small for an encryption key 🤔  Could we bump it to, say 128 if needed?

feature/search/src/main/kotlin/uk/gov/govuk/search/data/local/SearchDao.kt
```diff
@@ -0,0 +1,27 @@
...
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import kotlinx.coroutines.flow.Flow
+import uk.gov.govuk.search.data.local.model.LocalSearchItemEntity
+
+@Dao
+internal interface SearchDao {
+
+    @Query("SELECT * FROM local_search_items ORDER BY timestamp DESC")
```
gclssvglx, line 13: Are we controlling the DB schema at all, like is there a DDL to create the DB? I ask coz using `SELECT *` is not generally not recommended for 'un-schema'd' DBs. Basically, easy to hack a new field into the DB, enter something in there `DELETE * FROM...` type of thing, then this gets run when the `SELECT *` is called. If you select individual fields, and ensure they trap such things it's safer.

data/src/main/kotlin/uk/gov/govuk/data/local/RoomEncryptionHelper.kt
davidc-gds: Tink uses AES256_GCM under the hood which is 32 byte so not sure having a bigger blob in shared prefs would add anything?

feature/search/src/main/kotlin/uk/gov/govuk/search/data/local/SearchDao.kt
davidc-gds: This seems to be how Room works by default. You can probably specify the columns in the query but I'm not sure it's an issue. The schema is defined by the entities and I think room performs a check on startup to make sure the database structure matches the code. So if the user hacked another column in there it would just crash the app. I'm not sure it would actually execute any sql in the column anyway, it would either get mapped to the property in the object as a string or just ignored if there was no property for that column. Room queries are parameterised prepared statements so that also protects against SQL injection. I think all of this is super low risk as it would involve the user (or somebody who has the user's physical device) messing around with the app code and database file of a device they already have access to.

data/src/main/kotlin/uk/gov/govuk/data/local/RoomEncryptionHelper.kt
gclssvglx: Thanks - ok, fair enough

feature/search/src/main/kotlin/uk/gov/govuk/search/data/local/SearchDao.kt
gclssvglx: Yeah, it's all super edge-case. I know Room is a SQLite DB under the hood, so just thinking about how I'd normally secure one of those. Let's not worry for now.

feature/visited/src/main/kotlin/uk/gov/govuk/visited/data/store/VisitedLocalDataSource.kt
```diff
@@ -1,58 +1,32 @@
...
-    val visitedItems: Flow<List<VisitedItem>> = flow {
-        emitAll(
-            realmProvider.open().query<VisitedItem>().asFlow().map {
-                it.list.sortedByDescending { it.lastVisited }
-            }
-        )
-    }
+    val visitedItems: Flow<List<VisitedItemEntity>> get() = dao.getVisitedItems()
 
     suspend fun insertOrUpdate(title: String, url: String, lastVisited: LocalDateTime = LocalDateTime.now()) {
```
lamprosgk, line 17: Could use Room's @Upsert here

feature/visited/src/main/kotlin/uk/gov/govuk/visited/data/store/VisitedDatabase.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.govuk.visited.data.store
+
+import androidx.room.Database
+import androidx.room.RoomDatabase
+import uk.gov.govuk.visited.data.model.VisitedItemEntity
+
+@Database(entities = [VisitedItemEntity::class], version = 1, exportSchema = false)
```
lamprosgk: Might be good to export the schema, can help with auto-migration according to docs

feature/visited/src/main/kotlin/uk/gov/govuk/visited/di/VisitedModule.kt
```diff
@@ -1,16 +1,46 @@
...
+
+    @Provides
+    @Singleton
+    fun providesVisitedDatabase(
+        @ApplicationContext context: Context,
+        roomEncryptionHelper: RoomEncryptionHelper,
+        realmEncryptionHelper: RealmEncryptionHelper,
+        analyticsClient: AnalyticsClient
+    ): VisitedDatabase {
+        val key = roomEncryptionHelper.getKey()
```
lamprosgk, line 34: Would be good security-wise to clear the key from the memory 
([example](https://proandroiddev.com/how-to-encrypt-your-room-database-in-android-using-sqlcipher-0bce78328bd6))

feature/topics/src/main/kotlin/uk/gov/govuk/topics/data/local/TopicsLocalDataSource.kt
```diff
@@ -1,90 +1,59 @@
...
+    private val dao: TopicsDao,
     private val topicsDataStore: TopicsDataStore
 ) {
 
-    val topics: Flow<List<LocalTopicItem>> = flow {
-        emitAll(realmProvider.open().query<LocalTopicItem>().asFlow().map { it.list })
-    }
+    val topics: Flow<List<LocalTopicItemEntity>> get() = dao.getTopics()
 
     suspend fun sync(
```
lamprosgk, line 22: We could bundle the update and insert operations to a single transaction with Room's `database.withTransaction { }` I think 
(would need to inject the database to TopicsLocalDataSource)

feature/search/build.gradle.kts
```diff
@@ -42,11 +42,11 @@ sonar {
     properties {
         property(
             "sonar.coverage.exclusions",
-            properties["sonar.coverage.exclusions"].toString() + ",**/SearchRealmProvider.*"
+            properties["sonar.coverage.exclusions"].toString() + ",**/SearchDatabase.*,**/SearchMigrationCallback.*"
```
lamprosgk: Would be good to allow SearchMigrationCallback so that we have coverage for migration

feature/local/src/main/kotlin/uk/govuk/app/local/data/local/LocalMigrationCallback.kt
```diff
@@ -0,0 +1,53 @@
...
+import uk.govuk.app.local.data.local.model.StoredLocalAuthority
+import uk.govuk.app.local.data.local.model.StoredLocalAuthorityParent
+
+internal class LocalMigrationCallback(
+    private val realmEncryptionHelper: RealmEncryptionHelper,
+    private val analyticsClient: AnalyticsClient
+) : RoomDatabase.Callback() {
+
+    override fun onOpen(db: SupportSQLiteDatabase) {
+        val config = runBlocking {
```
lamprosgk: Is there a way to do this (and the other runBlocking calls) asynchronously?

lamprosgk:
Looks good, nice and clean 👌🏼

feature/local/src/main/kotlin/uk/govuk/app/local/data/local/LocalMigrationCallback.kt
davidc-gds: I'm not sure there is. We need to bridge the suspending getRealmKey into the non suspending onOpen. The opening of the db and running of the migration are synchronous operations. The DAO call which triggers the creating of the db will run on a background thread so I don't think it's an issue.

data/src/main/kotlin/uk/gov/govuk/data/local/RoomEncryptionHelper.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.google.crypto.tink.RegistryConfiguration
+import com.google.crypto.tink.aead.AeadConfig
+import com.google.crypto.tink.integration.android.AndroidKeysetManager
+import dagger.hilt.android.qualifiers.ApplicationContext
+import java.security.SecureRandom
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class RoomEncryptionHelper @Inject constructor(
```
TimGJD, line 11: There is already a `TinkClient `and `CryptoProvider` interface in `data` module which manages a master key, could we use that rather than duplicating here? 
On reading it sounds like with Tink, having 1 master key for an app is encouraged?

data/src/main/kotlin/uk/gov/govuk/data/local/RoomEncryptionHelper.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        val aead = AndroidKeysetManager.Builder()
+            .withSharedPref(context, KEYSET_NAME, PREFS_NAME)
+            .withKeyTemplate(KeyTemplates.get("AES256_GCM"))
+            .withMasterKeyUri(MASTER_KEY_URI)
+            .build()
+            .keysetHandle
+            .getPrimitive(RegistryConfiguration.get(), Aead::class.java)
+
+        val stored = prefs.getString(DB_KEY_PREF, null)
```
TimGJD, line 24: Just to mention, I saw [data store tink](https://developer.android.com/jetpack/androidx/releases/datastore#1.3.0-alpha07) which is new that we could consider to get away from shared prefs?  (still in alpha)

davidc-gds: Probably not a good idea at this stage with it being in alpha

feature/visited/src/main/kotlin/uk/gov/govuk/visited/data/store/VisitedLocalDataSource.kt
```diff
@@ -1,58 +1,29 @@
...
```
lamprosgk, line 17: This is getting leaked and `VisitedClient` is using it directly, the client should call the repo.

feature/visited/src/main/kotlin/uk/gov/govuk/visited/data/VisitedRepo.kt
lamprosgk, line 10: The repo should include `insertOrUpdate` too (also see VisitedLocalDataSource comment)

PR #694, opened 2026-04-30
GOVUKAPP-3467 Analytics config changes

TimGJD:
# Analytics config changes

- Only call onPageView() once when screens load
- Add reusable RememberLaunchedEffect() composable function which only ever runs once on initial composition

## JIRA ticket(s)
  - [GOVUKAPP-3467](https://govukverify.atlassian.net/browse/GOVUKAPP-3467)

### DIFF
--- a/app/src/main/kotlin/uk/gov/govuk/login/ui/BiometricSettingsScreen.kt
+++ b/app/src/main/kotlin/uk/gov/govuk/login/ui/BiometricSettingsScreen.kt
@@ -21,7 +21,7 @@ import uk.gov.govuk.R
 import uk.gov.govuk.design.ui.component.BodyRegularLabel
 import uk.gov.govuk.design.ui.component.ChildPageHeader
 import uk.gov.govuk.design.ui.component.LargeVerticalSpacer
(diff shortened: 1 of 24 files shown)


lamprosgk:
Could extract to a Composable to avoid repetition, for example 

```
@Composable
fun TrackScreenView(onPageView: () -> Unit) {
    var hasTrackedPageView by rememberSaveable { mutableStateOf(false) }

    LaunchedEffect(Unit) {
        if (!hasTrackedPageView) {
            onPageView()
            hasTrackedPageView = true
        }
    }
}
```

and use as

```
@Composable
fun SomeScreen(
    ...,
    onPageView: () -> Unit,  
) {
   
    TrackScreenView { onPageView() }
    ...
```

(If we need more functionality in a `LaunchedEffect` block could just add in a separate `LaunchedEffect` block)

TimGJD:
Was already coming up with something similar to [this ](https://github.com/govuk-once/govuk-mobile-android-app/pull/694#issuecomment-4353311058) when this PR was in draft. I had named things differently though as thought this could be useful for other things and not just for calls to onPageView()?

lamprosgk (approved):

PR #701, opened 2026-05-07
GOVUKAPP-3516 : Ts&Cs date parsing bug

gclssvglx:
## Summary
Fixed a bug where devices running **Android API 29 through 33** were stuck on the "App Unavailable" screen after login. This was caused by a compatibility issue with `java.time.Instant` when parsing ISO-8601 timestamps with offsets.

## The Issue
We compare the user's `accepted_at` timestamp against the API's `public_updated_at` (e.g., `2026-03-20T11:25:39+00:00`).

* **Observation:** While `java.time.Instant.parse()` works on modern Kotlin environments and Android 14+, it fails on older Android API levels when encountering specific offset strings.
* **Result:** The comparison logic fails, and the app defaults to the "App Unavailable" state.

## The Fix
Replaced `java.time.Instant` with `java.time.OffsetDateTime`. This class provides better compatibility for parsing zoned and offset timestamps across all targeted Android versions.

### Code Comparison
```kotlin
// Old approach: Fails on API < 34 for strings with offsets (+00:00)
Instant.parse(timestampOffset).toEpochMilli()

// New approach: Stable across all supported API levels
OffsetDateTime.parse(timestampOffset).toInstant().toEpochMilli()
```

| Android Version | API Level | Before Fix | After Fix |
| -- | -- | -- | -- |
| 10, 11, 12, 13 | 29 - 33 | ❌ Broken | ✅ Working |
| 14, 15, 16 | 34 - 36 | ✅ Working | ✅ Working |

## JIRA ticket(s)
  - [GOVUKAPP-3516](https://govukverify.atlassian.net/browse/GOVUKAPP-3516)

## Videos
| Before | After |
|--------|-------|
| [Screen_recording_20260507_094846.webm](https://github.com/user-attachments/assets/7fa52328-cb7a-497b-bf2c-907e745cc153) | [Screen_recording_20260507_094916.webm](https://github.com/user-attachments/assets/eaf2cd34-4a5d-47f3-aeda-4c9a55bb49e6) |

### DIFF
--- a/app/src/main/kotlin/uk/gov/govuk/terms/data/TermsRepo.kt
+++ b/app/src/main/kotlin/uk/gov/govuk/terms/data/TermsRepo.kt
@@ -1,5 +1,6 @@
 package uk.gov.govuk.terms.data
 
+import uk.gov.govuk.analytics.AnalyticsClient
 import uk.gov.govuk.config.data.ConfigRepo
 import uk.gov.govuk.terms.data.local.TermsDataStore
 import java.time.Instant
@@ -17,7 +18,8 @@ internal sealed class TermsAcceptanceState {
 @Singleton
 internal class TermsRepo @Inject constructor(
     private val termsDataStore: TermsDataStore,
-    private val configRepo: ConfigRepo
+    private val configRepo: ConfigRepo,
+    private val analyticsClient: AnalyticsClient
 ) {
     internal suspend fun termsAccepted(acceptedDate: Long = Instant.now().toEpochMilli()) {
         termsDataStore.setTermsAcceptedDate(acceptedDate)
@@ -34,7 +36,8 @@ internal class TermsRepo @Inject constructor(
             } else {
                 TermsAcceptanceState.Accepted
             }
-        } catch (_: Exception) {
+        } catch (e: Exception) {
+            analyticsClient.logException(e)
             TermsAcceptanceState.Error
         }
     }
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/uk/gov/govuk/terms/data/TermsRepo.kt
TimGJD, line 41: I wondered if we should/could add some remote logging here so we can see from logs if there's any future issues here as it blocks the app from being used when it fails?

gclssvglx: Yeah, good shout! As we currently don't really do any logging in the app, what do you recommend - [Timber](https://github.com/jakewharton/timber) seems like the Android default goto? Would need to add the dependency of course. Or the standard `Log.d` maybe?

lamprosgk: I think if we want to keep the repository pure best not use `Log.d` which is an Android sdk dependency. `Timber` (I think that's what you mean @gclssvglx) is much better but technically still not pure Kotlin. I've seen other projects creating a 'Logger' interface to avoid the android sdk dependency, I think that's what DI are doing too. 
(I think here though Tim means remote logging like Crashlytics/firebase)

gclssvglx: Thanks @lamprosgk 👍  Yes, I was thinking of Timber - but @TimGJD suggests we use the existing Firebase crashlytics `logException` so I'm working on that.

TimGJD (approved):

PR #709, opened 2026-05-13
GOVUKAPP-3370 : Qualtrics integration

gclssvglx:
This PR implements a centralised integration of the Qualtrics SDK. The goal is to decouple Qualtrics in the app in a way that (once implemented) the Analytics team can create and manage surveys and feedback forms via the Experience Manager without any further developer intervention.

## JIRA ticket(s)
  - [GOVUKAPP-3370](https://govukverify.atlassian.net/browse/GOVUKAPP-3370)
  - [GOVUKAPP-3782](https://govukverify.atlassian.net/browse/GOVUKAPP-3782)

## Additions and updates to analytics events

Added `fb_session_id` and `fb_user_pseudo_id` to all existing 'regular' and ecommerce events...

```shell
Logging event: origin=app+gtm,name=Navigation,params=Bundle[{external=0, fb_session_id=1784816277, language=en, ga_event_origin(_o)=app+gtm, ga_screen_class(_sc)=HomeScreen, ga_screen_id(_si)=-2738979775568515666, ga_screen(_sn)=Homepage, text=Settings, type=Tab, fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1}]

Logging event: origin=app,name=screen_view(_vs),params=Bundle[{fb_session_id=1784816277, language=en, screen_title=Settings, ga_event_origin(_o)=app, engagement_time_msec(_et)=49933, ga_previous_class(_pc)=HomeScreen, ga_previous_id(_pi)=-2738979775568515666, ga_previous_screen(_pn)=Homepage, ga_screen_class(_sc)=SettingsScreen, ga_screen_id(_si)=-2738979775568515665, ga_screen(_sn)=Settings, manual_tracking(_mst)=1, fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1}]

Logging event: origin=app+gtm,name=view_item_list,params=Bundle[{item_list_name=Your topics, fb_session_id=1784816277, item_list_id=Your topics, ga_event_origin(_o)=app+gtm, ga_screen_class(_sc)=HomeScreen, ga_screen_id(_si)=-2738979775568515664, ga_screen(_sn)=Homepage, items=[Bundle[{item_name=Driving and transport, item_list_name=Your topics, item_list_id=Your topics, index=1, location_id=driving-transport, results=1}]], fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1}]
```

Added two new events `qualtrics_survey_shown` and  `qualtrics_survey_closed` which (in addition to the regular data fields) also have `fb_session_id`, `fb_user_pseudo_id` and `qualtrics_targeting_id`. **Note**: if there is more than 1 `qualtrics_targeting_id` returned you'll get one event for each `qualtrics_targeting_id`.

```shell
Logging event: origin=app+gtm,name=qualtrics_survey_shown,params=Bundle[{fb_session_id=1784816277, qualtrics_targeting_id=SI_8nRe24gWo6vCvBQ, ga_event_origin(_o)=app+gtm, ga_screen_class(_sc)=SettingsScreen, ga_screen_id(_si)=-2738979775568515665, ga_screen(_sn)=Settings, fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1}]

Logging event: origin=app+gtm,name=qualtrics_survey_closed,params=Bundle[{fb_session_id=1784816277, qualtrics_targeting_id=SI_8nRe24gWo6vCvBQ, ga_event_origin(_o)=app+gtm, ga_screen_class(_sc)=SettingsScreen, ga_screen_id(_si)=-2738979775568515665, ga_screen(_sn)=Settings, fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1}]
```


analytics/src/main/kotlin/uk/gov/govuk/analytics/FirebaseAnalyticsClient.kt
```diff
@@ -1,18 +1,23 @@
...
+import com.qualtrics.digital.Qualtrics
+import dagger.hilt.android.qualifiers.ApplicationContext
 import uk.gov.govuk.analytics.data.local.model.EcommerceEvent
 import java.io.Serializable
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class FirebaseAnalyticsClient @Inject constructor(
+    @ApplicationContext private val context: Context,
```
lamprosgk: AnalyticsClient should be a pure Kotlin class and have no knowledge of the Android framework (context)

gclssvglx: Hi @lamprosgk 👋  I completely agree with the principle of keeping AnalyticsClient a pure Kotlin entity for testability and separation of concerns. However, we face a specific constraint with the Qualtrics SDK: unlike standard data-only analytics (like Firebase), Qualtrics is a UI-driven tool. It requires a Context reference to evaluate project targeting and to anchor its survey View layers.

To bridge this while maintaining as much architectural 'purity' as possible, I considered two paths:
- Global Lifecycle Injection: Having a process-level observer handle the Context. I’ve avoided this as it’s too broad and makes user-consent scoping more difficult.
- Abstracted Context Provider: We can provide the Context to the Qualtrics implementation via a constructor-injected provider or a specialised wrapper that implements our pure-Kotlin analytics interface.

The current implementation ensures that while the implementation knows about the Android Framework, the call sites across the app remain pure and platform-agnostic. While I'm open to further decoupling this via a wrapper if we feel the Context dependency is too leaky for the current AnalyticsClient abstraction, I feel doing that will trigger app-wide changes. Something I'm seeking to avoid.

analytics/src/main/kotlin/uk/gov/govuk/analytics/FirebaseAnalyticsClient.kt
```diff
@@ -1,18 +1,23 @@
...
 import com.google.firebase.crashlytics.FirebaseCrashlytics
+import com.qualtrics.digital.Qualtrics
+import dagger.hilt.android.qualifiers.ApplicationContext
 import uk.gov.govuk.analytics.data.local.model.EcommerceEvent
 import java.io.Serializable
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class FirebaseAnalyticsClient @Inject constructor(
```
lamprosgk, line 13: This tightly couples Firebase and Qualtrics together in an 'firebase' client. Ideally they should be abstracted, if the same events need to be tracked there should be an interface contract and an orchestrator class

gclssvglx: Thanks @lamprosgk 👍 I've pretty much done as you suggest. However, there are also other changes relating to an issue I found in the Qualtrics SDK - basically it caches events (or doesn't clean them out once they are sent) - so data is 'leaked' across events. We're seeking advice from Qualtrics on this to find a better solution. But, for now - a list of keys which we either set or clear before sending is the only way.

Hope it's "Uncle Bob" enough now 😉

analytics/src/main/kotlin/uk/gov/govuk/analytics/QualtricsAnalyticsClient.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import android.content.Context
+import com.google.firebase.analytics.FirebaseAnalytics
+import com.qualtrics.digital.Qualtrics
+import dagger.hilt.android.qualifiers.ApplicationContext
+import uk.gov.govuk.analytics.data.local.model.EcommerceEvent
+import javax.inject.Inject
+
+class QualtricsAnalyticsClient @Inject constructor(
+    @ApplicationContext private val context: Context,
```
TimGJD: Android Studio is likely to have highlighted this to recommend adding a site-target? Looks like it would be `@param` in this instance

app/src/main/kotlin/uk/gov/govuk/GovUkApplication.kt
```diff
@@ -9,10 +10,13 @@ import javax.inject.Inject
 class GovUkApplication: Application() {
 
     @Inject lateinit var notificationsProvider: NotificationsProvider
+    @Inject lateinit var qualtricsProvider: Qualtrics
```
TimGJD: Wondering if it would be better if we had the `analytics coordinator interface/provider` here instead with an `initialise()` function called in `onCreate()` which calls `initializeProject()` in the `QualtricsAnalyticsClient`?

We could also then move those Qualtric constants into Analytics module too

analytics/src/main/kotlin/uk/gov/govuk/analytics/QualtricsAnalyticsClient.kt
gclssvglx: Thanks @TimGJD - I was not aware that `@param:ApplicationContext` is a thing. I had to do some Googleing on it. 

But, question - it seems `@ApplicationContext` is the 'modern' way of doing this as Kotlin will simply figure it out. Whereas, `@param:ApplicationContext` is classed as 'legacy' and 'appropriate for older codebases'. Should we not be shooting for modern over legacy?

app/src/main/kotlin/uk/gov/govuk/GovUkApplication.kt
gclssvglx: Thanks @TimGJD - I'll give this a try.

analytics/src/main/kotlin/uk/gov/govuk/analytics/QualtricsAnalyticsClient.kt
gclssvglx: Fixed up

app/src/main/kotlin/uk/gov/govuk/GovUkApplication.kt
gclssvglx: Thanks @TimGJD - great suggestion. However, I didn't like the `initialize()` function in the `AnalyticsCoordinator` - seems like it then knew too much about the `QualtricsAnalyticsClient`. So, I simply created the `initialize()` method in the `QualtricsAnalyticsClient` and called that in the `onCreate()`. A provider then wasn't required.

analytics/src/main/kotlin/uk/gov/govuk/analytics/QualtricsAnalyticsClient.kt
```diff
@@ -9,9 +9,17 @@ import javax.inject.Inject
 
 class QualtricsAnalyticsClient @Inject constructor(
     @param:ApplicationContext private val context: Context,
-    private val qualtrics: Qualtrics
+    private val qualtrics: Qualtrics = Qualtrics.instance()
```
TimGJD: Do we need to assign this here? I think the Qualtrics provider should initialise this for us?

app/src/main/kotlin/uk/gov/govuk/GovUkApplication.kt
TimGJD: My thoughts were that `GovUkApplication` is closely coupled to `QualtricsAnalyticsClient`. 1 reason for using the interface would have been that if we ever wanted to swap out Qualtrics for something else we wouldn't need to change `GovUkApplication`. Like how we use NotificationsProvider so if we ever want to swap out One Signal for a different notifications provider the changes would be few as possible.

analytics/src/main/kotlin/uk/gov/govuk/analytics/QualtricsAnalyticsClient.kt
gclssvglx: Yes, I would have thought that but it didn't seem to. I'll look into it.

analytics/src/main/kotlin/uk/gov/govuk/analytics/FirebaseAnalyticsClient.kt
lamprosgk: Looks much better, good job!

analytics/src/main/kotlin/uk/gov/govuk/analytics/di/AnalyticsModule.kt
```diff
@@ -32,6 +36,25 @@ internal class AnalyticsModule {
     @Singleton
     fun provideCrashlytics(): FirebaseCrashlytics = Firebase.crashlytics
 
+    @Provides
+    @Singleton
+    fun provideQualtrics(): Qualtrics = Qualtrics.instance()
+
+    @Provides
+    @Singleton
+    fun provideAnalyticsCoordinator(@ApplicationContext context: Context): AnalyticsCoordinator {
```
TimGJD: Should the return type for this function be `AnalyticsCoordinatorInterface`?

analytics/src/main/kotlin/uk/gov/govuk/analytics/FirebaseAnalyticsClient.kt
lamprosgk: I think we are passing the wrong type of context though. To anchor to View layer Activity-level context is needed, the injected via Hilt ApplicationContext can cause a crash.
A way to do it without app wide changes (I agree) could be instead of calling `qualtrics.display(context)` in the analytics client, is to have a flow emitting a flag and listen to it in the activity and call `qualtrics.display(context)` there with the activity context.

gclssvglx:
Superceeded by https://github.com/govuk-once/govuk-mobile-android-app/pull/791

PR #715, opened 2026-05-19
GOVUKAPP-3568 Problem Screen - hide icon when landscaped

TimGJD:
# Problem Screen - hide icon when landscaped

- Hide the problem screen icon when device orientation is landscaped. 
- Reuse problem screen in Chat feature.

## JIRA ticket(s)
  - [GOVUKAPP-3568](https://govukverify.atlassian.net/browse/GOVUKAPP-3568)

## Screen shots

| Before | After |
|--------|-------|
| Screenshot_1779188734 | Screenshot_1779188855 |

| Before | After |
|--------|-------|
| Screenshot_1779182009 | Screenshot_1779184197 |

| Before | After |
|--------|-------|
| Screenshot_1779188168 | Screenshot_1779186884 |

| Before | After |
|--------|-------|
| Screenshot_1779188211 | Screenshot_1779186753 |

| Before | After |
|--------|-------|
| Screenshot_1779188234 | Screenshot_1779187050 |


design/src/main/kotlin/uk/gov/govuk/design/ui/component/error/ErrorPage.kt
```diff
@@ -29,26 +31,30 @@ import uk.gov.govuk.design.ui.theme.GovUkTheme
 @Composable
 fun ErrorPage(
     headerText: String,
-    subText: String,
-    footer: @Composable () -> Unit,
+    vararg subText: String,
```
lamprosgk: While this is technically correct, I saw the recommendation for Compose is to use a list instead (reason being that `vararg` is basically an array which is mutable, and can force unnecessary recompositions - some documentation [here](https://developer.android.com/develop/ui/compose/performance/stability) ). May be better to use a list.

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/component/ErrorScreens.kt
```diff
@@ -1,106 +1,49 @@
...
-        additionalText = stringResource(id = R.string.error_retry_page_additional_text),
-        buttonText = stringResource(id = R.string.error_retry_button_text),
-        onRetry = onRetry
+    ErrorPage(
+        headerText = stringResource(id = R.string.error_page_header),
+        subText = arrayOf(
+            stringResource(id = R.string.error_retry_page_sub_text_1),
+            stringResource(id = R.string.error_retry_page_sub_text_2)
+        ),
+        modifier = modifier,
```
lamprosgk, line 26: Do we not need the `padding(horizontal = GovUkTheme.spacing.medium)` padding here?

lamprosgk (approved):

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/component/ErrorScreens.kt
TimGJD: I've checked the Chat [Figma](https://www.figma.com/design/rKYEySuNc8aikJWYfEz1wR/2025-04-GOV.UK-Chat?node-id=6079-36910&t=cJmmiP6mWcZI9BHD-0) and can see the 'problem screen' is supposed to have 16px padding all round. 

The current horizontal padding in main branch is 32px (which is noticeable in the first 'before' screen shot in this PR description). This is because 16px horizontal padding was added here AND the 'central aligned screen' component used for this screen also adds 16px padding. 

My change of removing the horizontal padding now aligns the padding with the Figma designs.

lamprosgk: Thanks for checking 👍🏼

PR #718, opened 2026-05-21
GOVUKAPP-3590 Dvla linking failure tracking

lamprosgk:
# Title

Added analytics for linking failure cases.

## JIRA ticket(s)
  - [GOVUKAPP-3590](https://govukverify.atlassian.net/browse/GOVUKAPP-3590)


feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/DvlaLinkingRoute.kt
```diff
@@ -100,16 +100,29 @@ internal fun DvlaLinkingRoute(
...
         is DvlaViewModel.UiState.Error.Other -> {
-            ErrorOtherScreen(
-                onWebFlowClosed = onWebFlowClosed,
-                onLaunchBrowser = onLaunchBrowser,
+            DvlaLinkErrorScreen(
+                onPageView = { title ->
+                    viewModel.onErrorOtherPageView(title)
+                },
+                onBackToDrivingClicked = { buttonText ->
+                    viewModel.onErrorBackToDrivingClicked(buttonText)
```
TimGJD, line 119: Noticing here that function calls with only 1 argument, the argument is not named. Function calls with more than 1 argument then the arguments are named. 

Just checking this is intentional and is a pattern for all to follow?

lamprosgk: Is there a pattern we follow already for one arg calls? Happy to change if there is.

TimGJD: Not that I'm aware of and it seems elsewhere we sometimes name the arg on one arg calls and sometimes not. 

Was just wondering so we know wether to comment these or not.

TimGJD (approved):

PR #723, opened 2026-05-28
GOVUKAPP-3545 Use segmented control on vehicle and licence summary widget

TimGJD:
# Use segmented control on vehicle and licence summary widget

## JIRA ticket(s)
  - [GOVUKAPP-3545](https://govukverify.atlassian.net/browse/GOVUKAPP-3545)

## Figma
  - [Figma board](https://www.figma.com/design/utmaFnJPNY4sVgwjui5MmK/Driving?node-id=5449-38421&t=0EbmJ9MccAn0QASH-0)

## Screen shots

| Light | Dark |
|--------|-------|
| Screenshot_1779982180 | Screenshot_1779982214 |


feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/model/VehicleAndLicenceSummaryWidgetUiState.kt
```diff
@@ -0,0 +1,22 @@
+package uk.gov.govuk.dvla.ui.model
+
+internal sealed interface UiState {
+    data class Default(val category: Category) : UiState
+    data object Hidden : UiState
+}
+
+enum class Category {
```
lamprosgk: This is probably an inherited name from the Topics segmented control, I think something driving related would be more suitable here

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/data/local/DvlaDataStore.kt
```diff
@@ -0,0 +1,38 @@
...
+import javax.inject.Inject
+import javax.inject.Named
+import javax.inject.Singleton
+
+@Singleton
+class DvlaDataStore @Inject constructor(
+    @param:Named("dvla_prefs") private val dataStore: DataStore<Preferences>
+) {
+    internal companion object {
+        const val SELECTED_CATEGORY = "selected_category"
```
lamprosgk: Same as the `VehicleAndLicenceSummaryWidgetUiState`, I think 'category' name is too generic

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/VehicleAndLicenceSummaryViewModel.kt
```diff
@@ -9,34 +9,36 @@ import kotlinx.coroutines.launch
...
 internal class VehicleAndLicenceSummaryViewModel @Inject constructor(
     private val dvlaRepo: DvlaRepo,
     private val mapper: VehicleSummaryMapper
 ) : ViewModel() {
 
-    private val _uiState = MutableStateFlow(
-        if (dvlaRepo.linkState.value == DvlaLinkState.LINKED) VehicleAndLicenceSummaryUiState.Loading
-        else VehicleAndLicenceSummaryUiState.Hidden
-    )
+    private val _uiState = MutableStateFlow<UiState?>(null)
```
lamprosgk: It's better to avoid a nullable state/initialise with null and have a 'loading' state or similar.

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/VehicleAndLicenceSummaryWidget.kt
```diff
@@ -6,33 +6,102 @@ import androidx.compose.foundation.layout.fillMaxWidth
...
+
+                    MediumVerticalSpacer()
 
-    when (val currentState = state) {
-        is VehicleAndLicenceSummaryUiState.Hidden -> return // draw nothing if not linked
-        is VehicleAndLicenceSummaryUiState.Loading -> VehicleAndLicenceSummaryLoading(modifier = modifier)
-        is VehicleAndLicenceSummaryUiState.Error -> {
+                    when (it.category) {
+                        Category.VEHICLE -> {
+                            activeButtonState = VehicleButton
```
lamprosgk: This is mutating the state and during composition and it's an antipattern, can create a recomposition loop.
rememberSaveable isn't needed, the similar `when` block it `TopicsWidget` is a good example:

```
val activeButtonState = when (uiState.selectedCategory) {
    TopicsCategory.YOUR -> ConnectedButton.FIRST
    TopicsCategory.ALL -> ConnectedButton.SECOND
}
```

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/VehicleAndLicenceSummaryWidget.kt
```diff
@@ -6,33 +6,102 @@ import androidx.compose.foundation.layout.fillMaxWidth
...
 fun VehicleAndLicenceSummaryWidget(
-    modifier: Modifier = Modifier,
+    modifier: Modifier = Modifier
 ) {
     val viewModel: VehicleAndLicenceSummaryViewModel = hiltViewModel()
-    val state by viewModel.uiState.collectAsState()
+    val uiState by viewModel.uiState.collectAsState()
+    val vehicleSummaryUiState by viewModel.vehicleSummaryUiState.collectAsState()
+    val licenceSummaryUiState by viewModel.licenceSummaryUiState.collectAsState()
+    var activeButtonState by rememberSaveable { mutableStateOf(VehicleButton) }
```
lamprosgk: (see below)

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/model/VehicleAndLicenceSummaryWidgetUiState.kt
TimGJD: I have a block when thinking of what to call this so 'category' is all I could come up with. 
@lamprosgk you will need to give me a better name suggestion than category in order for this to be changed.

lamprosgk: Maybe something like DrivingSection/DrivingView?

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/VehicleAndLicenceSummaryViewModel.kt
TimGJD: Can you give a strong argument for this as there are 12 other occurrences in the app where we have set mutable state initial value as `null` so I was following usual practice?

lamprosgk: The main reason is that null doesn't actually represent a distinct state (like say 'loading'), so in UDF terms we are passing Compose a state which doesn't actually mean anything/it doesn't know how to draw. Also, when the state is non nullable sealed interface, we are forced to have an exhaustive when block in Compose (which guarantees no state will be missed).

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/VehicleAndLicenceSummaryWidget.kt
TimGJD: Ah yes, thats way better, only 1 recomposition on button press now instead of unnecessary 2, thanks

lamprosgk (approved):

PR #726, opened 2026-06-01
GOVUKAPP-3608 Licence view summary card

lamprosgk:
# Title

- Added Licence card, following the same pattern as Vehicles 
- Simplified vehicles/licence composables to be reusable

## JIRA ticket(s)
  - [GOVUKAPP-3608](https://govukverify.atlassian.net/browse/GOVUKAPP-3608)

## Figma
  - [Figma](https://www.figma.com/design/utmaFnJPNY4sVgwjui5MmK/Driving?node-id=6501-49710&t=H7WtJ2M8tAdRXjzG-1)

## Screen shots
[Screen_recording_20260605_120210.webm](https://github.com/user-attachments/assets/2d938a22-bfd8-409c-ad89-fc9aa0b4f94e)

Screenshot_20260605_120248
Screenshot_20260605_120255


design/src/main/kotlin/uk/gov/govuk/design/ui/component/List.kt
```diff
@@ -335,13 +335,14 @@ fun IconListItem(
...
+    modifier: Modifier = Modifier,
+    title: String? = null,
     description: String,
     @DrawableRes icon: Int?,
-    modifier: Modifier = Modifier,
     isFirst: Boolean = false,
     isLast: Boolean = false,
-    titleAltText: String? = null
+    titleAltText: String? = null,
+    descriptionAltText: String? = null
```
TimGJD: Would a data class with text and altText constructor params for title and description be cleaner here?

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/VehiclesAndLicenceSummaryWidget.kt
```diff
@@ -31,44 +40,64 @@ fun VehiclesAndLicenceSummaryWidget(
...
-                                LicenceButton -> viewModel.onLicenceSelected()
-                            }
-                        },
-                        colours = ButtonColours(
-                            containerActive = GovUkTheme.colourScheme.surfaces.connectedButtonGroupActive,
-                            containerInactive = GovUkTheme.colourScheme.surfaces.list
-                        )
+    val context = LocalContext.current
+    val hapticFeedback = LocalHapticFeedback.current
+    val licenceClipboardLabel = stringResource(R.string.clipboard_data_label_licence_number)
```
TimGJD: Move context, hapticFeedback, licenceClipboardLabel val's into the scope where they are used as they appear to be only read if the state is LICENCE?

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/util/DateExtension.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.govuk.dvla.util
+
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+
+private val dvlaDateFormatter = DateTimeFormatter.ofPattern("d MMMM yyyy")
+
+fun LocalDate.toSummaryDisplayFormat(): String = this.format(dvlaDateFormatter)
```
TimGJD, line 8: Would a `runCatching()` be a good idea here as it looks like we could be relying on DVLA API suppling us date in specific format which they control so could therefore change?

feature/dvla/src/test/java/uk/gov/govuk/dvla/domain/DriverSummaryMapperTest.kt
```diff
@@ -5,48 +5,81 @@ import io.mockk.mockk
 import org.junit.Assert.*
```
TimGJD: I know this not in changes but remove this wildcard whilst we're here?

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/VehiclesAndLicenceSummaryWidget.kt
```diff
@@ -77,48 +106,61 @@ fun VehiclesAndLicenceSummaryWidget(
...
+                licenceSummary = licenceState.licence,
+                onMoreClick = { /* TODO to be handled in next ticket(s) */ },
+                onLicenceNumberLongClick = { onLicenceNumberLongClick(licenceState.licence.licenceNumber) },
+                modifier = modifier
+            )
         }
     }
 }
 
+private fun copyToClipboard(
```
TimGJD: Could be String extension function here instead of `textToCopy` param to reduce function params?

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/util/DateExtension.kt
lamprosgk: That wouldn't be needed here as we're converting a date object (would need if it was a String, which we do in `toLocalDateOrNull()` extension function where it's converted to the LocalDate used here).

design/src/main/kotlin/uk/gov/govuk/design/ui/component/List.kt
lamprosgk: Have added, it is cleaner here but I think it's become more verbose when calling the composable

TimGJD (approved):

PR #739, opened 2026-06-16
GOVUKAPP-3555 Add vehicle card

lamprosgk:
# Title

Added cards to add vehicles when customer has no vehicles/add a vehicles when there are vehicles. 

To try, link the following customer ids:
`8bc39b9f-9cdc-4dc6-8364-632d1f7b5916` has no vehicles.
`49fec1c5-0712-4a2d-9116-406eb147ff7a` has vehicles.

## JIRA ticket(s)
  - [GOVUKAPP-3555](https://govukverify.atlassian.net/browse/GOVUKAPP-3555)

## Figma
  - [layout](https://www.figma.com/design/utmaFnJPNY4sVgwjui5MmK/Driving?node-id=5424-28889&t=Hbl1PUGnTi7mdaRd-1)
  - [documentation](https://www.figma.com/design/utmaFnJPNY4sVgwjui5MmK/Driving?node-id=5449-38393&t=Hbl1PUGnTi7mdaRd-1)

## Screen shots
Screenshot_20260616_141259
Screenshot_20260616_142803


design/src/main/kotlin/uk/gov/govuk/design/ui/component/Card.kt
```diff
@@ -335,24 +336,30 @@ fun CentredCardWithIcon(
     @DrawableRes icon: Int,
     modifier: Modifier = Modifier,
     title: String? = null,
-    description: String? = null
+    description: String? = null,
+    drawBottomStroke: Boolean = true,
+    verticalPadding: Dp = GovUkTheme.spacing.extraLarge
```
TimGJD: Just wondering if having a modifier here instead for the centred content would be good rather than specific vertical padding incase further design requirements in future?

lamprosgk: That's a good point, but having a second modifier is not a good idea I think, I will change to `PaddingValues` instead which will make it lot more flexible 👍🏼

TimGJD (dismissed):

davidc-gds (approved):

PR #749, opened 2026-06-29
GOVUKAPP-3698 : Error screen when chat window is backgrounded

gclssvglx:
# Error screen when chat window is backgrounded

When a user types a question, and the app is in the process of generating an answer - if the app is backgrounded by the user, an error screen is shown indicating that there is a problem with the chat screen. 

Essentially, the POST question API call is made and Chat starts 'polling' GET API calls for the answer. If the answer is retrieved while the app is still in background - irrespective of the `battery allowed in background` setting (API v15+) - the answer cannot be displayed.

However, if the app is still waiting for the answer to be generated - when the app is foregrounded, Chat behaves as normal and simply displays the answer when it is successfully retrieved.

To 'fix' the confused `error page` state, the user can simply click 'Home' or 'Settings', then 'Chat' again - the error will clear and the answer will be shown as usual via `loadConversation`. It is this that gives us a clue to a potential - more suitable fix.

The approach taken is to identify when the app is backgrounded - using the lifecycle `onPause` event. This will stop any on-going API calls (e.g. polling) and clean up the `uiState`. Then when the app is foregrounded - using the lifecycle `onResume` event. The `loadConversation` function is called mimicking the behaviour as if the user had loaded Chat - effectively reloading the conversation.

GOVUKAPP-3698

## JIRA ticket(s)
  - [GOVUKAPP-3698](https://govukverify.atlassian.net/browse/GOVUKAPP-3698)

## Videos

| Setting | Before | After |
|--------|--------|-------|
| Screenshot_20260629_090020 | [Screen_recording_20260629_090112.webm](https://github.com/user-attachments/assets/cab94f5f-552c-4e30-86dc-0139761e4bfd) |   |
|Screenshot_20260629_090314 | [Screen_recording_20260629_090414.webm](https://github.com/user-attachments/assets/074c134d-f19d-4b33-85e9-84f8b68948ee) |   |

### DIFF
--- a/feature/chat/src/main/kotlin/uk/gov/govuk/chat/ChatViewModel.kt
+++ b/feature/chat/src/main/kotlin/uk/gov/govuk/chat/ChatViewModel.kt
@@ -3,6 +3,7 @@ package uk.gov.govuk.chat
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import dagger.hilt.android.lifecycle.HiltViewModel
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharedFlow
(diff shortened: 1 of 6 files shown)


feature/chat/src/main/kotlin/uk/gov/govuk/chat/ChatViewModel.kt
```diff
@@ -68,14 +69,46 @@ internal class ChatViewModel @Inject constructor(
...
+        }
+    }
+
+    private var loadConversationJob: Job? = null
+    private var askQuestionJob: Job? = null
+
+    fun loadConversation() {
+        if (loadConversationJob?.isActive == true) return
+
+        val currentState = _uiState.value
```
TimGJD: The below might be more simple if it works? 

```
 _uiState.value = when(val currentState = _uiState.value) {
            is Default ->  currentState.copy(isLoading = true)
            else -> Default(isLoading = true)
        }
```

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ChatViewModel.kt
```diff
@@ -68,14 +69,46 @@ internal class ChatViewModel @Inject constructor(
...
+        loadConversationJob?.cancel()
+        askQuestionJob?.cancel()
+        val currentState = _uiState.value
+        if (currentState is Default) {
+            _uiState.value = currentState.copy(isLoading = false)
+        }
+    }
+
+    private var loadConversationJob: Job? = null
+    private var askQuestionJob: Job? = null
```
TimGJD: I think our usual order is to declare any variables before any functions?

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ChatViewModel.kt
davidc-gds: Fixed

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ChatViewModel.kt
davidc-gds: Fixed

TimGJD (approved):

PR #756, opened 2026-07-01
Notification centre

mattmapadmi:
# Notification Centre

This is several months of work, including the original PoC code.

High level features are:
* Link to Messages in Settings (which checks if you have a DVLA account linked, otherwise hides it)
* Message list
* Message detail, with Markdown rendering

## JIRA ticket(s)
  - [NOT-237](https://gdsgovukagents.atlassian.net/browse/NOT-237)
  - [NOT-224](https://gdsgovukagents.atlassian.net/browse/NOT-224)
  - [NOT-226](https://gdsgovukagents.atlassian.net/browse/NOT-226)

## Figma
  - https://www.figma.com/design/352H6yuUPABtuZjdeXEyp0/Settings?node-id=31-8303&m=dev

## Screenshots

Screenshot_20260715_212447
Screenshot_20260715_212458
Screenshot_20260715_212504
Screenshot_20260715_212513

[NOT-237]: https://gdsgovukagents.atlassian.net/browse/NOT-237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NOT-224]: https://gdsgovukagents.atlassian.net/browse/NOT-224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NOT-226]: https://gdsgovukagents.atlassian.net/browse/NOT-226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/main/kotlin/uk/gov/govuk/navigation/TopLevelDestination.kt
```diff
@@ -27,7 +30,9 @@ internal sealed class TopLevelDestination(
         associatedRoutes = listOf(
             VISITED_ROUTE,
             TOPIC_ROUTE,
-            TOPICS_ALL_STEP_BY_STEPS_ROUTE
+            TOPICS_ALL_STEP_BY_STEPS_ROUTE,
+            NOTIFICATION_CENTRE_GRAPH_START_DESTINATION,
```
davidc-gds: Should these routes be associated with setting rather than home if the notification centre is accessed via settings?

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/NotificationCentreViewModel.kt
```diff
@@ -0,0 +1,80 @@
...
+
+            val notifications = notificationCentreRepo.getNotifications()
+            withContext(Dispatchers.Main) {
+                _uiState.value = when (notifications) {
+                    is Result.Success -> {
+                        if (notifications.value.isEmpty()) {
+                            NotificationCentreUiState.Empty
+                        } else {
+                            val sorted = notifications.value.sortedBy { it.date }
+                            val sevenDaysBack = Instant.now().minus(7, ChronoUnit.HOURS)
```
davidc-gds: This looks like it's using the wrong unit (hours rather than days), won't this impact messages older than 7 hours rather than 7 days?

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/ui/NotificationCentreDetailScreen.kt
```diff
@@ -0,0 +1,489 @@
...
+    modifier: Modifier = Modifier,
+    actionColour: Color = GovUkTheme.colourScheme.textAndIcons.linkHeader,
+    onBack: () -> Unit,
+    onUnread: () -> Unit,
+    onDelete: () -> Unit
+) {
+    Column(
+        modifier
+            .background(GovUkTheme.colourScheme.surfaces.homeHeader)
+            .semantics { this.hideFromAccessibility() }
```
davidc-gds: Won't applying this to the entire column impact talkback for all of it's children i.e. back, unread and delete buttons? Is this intended behaviour?

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/ui/NotificationCentreDetailScreen.kt
```diff
@@ -0,0 +1,489 @@
...
+                )
+                
+                CompositionLocalProvider(
+                    LocalUriHandler provides object : UriHandler {
+                        override fun openUri(uri: String) {
+                            launchBrowser(uri)
+                        }
+                    }
+                ) {
+                    Markdown(
```
davidc-gds: It might be worth looking at the custom markdown rendering we do in the chat module as that addressed some issue we had with other third party markdown rendering (specifically around accessibility if I remember correctly).

davidc-gds (commented):
Module coupling: settings → notificationcentre

Something to think about/discuss. We want to try and limit coupling as much as possible between the features. At the moment settings and notification centre are quite tightly coupled. There are examples of how we handle feature to feature dependency elsewhere in the app e.g. search/topics and visited. We do this via a lightweight feature interface and controlling the visibility of the feature's classes. Could we look at doing something similar here?

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/ui/NotificationCentreDetailScreen.kt
mattmapadmi: Not sure what happened there, wasn't the intended behaviour.

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/ui/NotificationCentreDetailScreen.kt
mattmapadmi: Ah, had a feeling there would've been a good reason for not using the MikePenz library. I'd added it originally to do a spike and generate some screenshots, and then meant to reassess closer to production. Have removed now and following Chat's lead.

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/NotificationCentreViewModel.kt
mattmapadmi: Good catch 🤦 Have improved the tests as well as they should've caught this.

app/src/main/kotlin/uk/gov/govuk/navigation/TopLevelDestination.kt
mattmapadmi: Ah yeah, it started life in Home and in the moving I'd forgotten to adjust the navigation.

mattmapadmi:
> Module coupling: settings → notificationcentre
> 
> Something to think about/discuss. We want to try and limit coupling as much as possible between the features. At the moment settings and notification centre are quite tightly coupled. There are examples of how we handle feature to feature dependency elsewhere in the app e.g. search/topics and visited. We do this via a lightweight feature interface and controlling the visibility of the feature's classes. Could we look at doing something similar here?

That makes sense. Have decoupled it with a thin shim as Settings only needed the UnreadCount anyway.

app/src/main/kotlin/uk/gov/govuk/navigation/DeeplinkHandler.kt
```diff
@@ -68,12 +69,17 @@ internal class DeeplinkHandler @Inject constructor(
     fun handleDeeplink(navController: NavController) {
         deepLink?.let {
 
-            // check for intercepted route first
-            if (interceptLinkedServiceCallback(it, navController)) {
+            // check for intercepted routes first
+            if (interceptOldDvlaAuthCallback(it, navController) ||
```
davidc-gds: DVLA appear to have deployed their migration to the new deeplink structure so this shouldn't be needed any more.

mattmapadmi: Ye was anticipating it wouldn't stay in this branch for long. Have removed it.

design/src/androidTest/kotlin/uk/gov/govuk/design/markdown/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
+    @Test
+    fun useAppContext() {
+        // Context of the app under test.
+        val appContext = InstrumentationRegistry.getInstrumentation().targetContext
+        assertEquals("uk.govuk.app.local.test", appContext.packageName)
```
TimGJD: I know we added this to just satisfy Sonar I think? Anyway, this will fail if we run Android tests as expected package name not correct I think?

mattmapadmi: 🤦 I copied this from somewhere and didn't notice it was more than an empty test.

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/ui/NotificationCentreDetailScreen.kt
```diff
@@ -0,0 +1,405 @@
...
+
+        if (showDeleteConfirmationDialog) {
+
+            ConfirmationDialog(
+                onConfirm = onConfirmDelete,
+                onCancel = onCancelDelete
+            )
+        }
+
+        LaunchedEffect(Unit) {
```
TimGJD: There is a 'RunOnceLaunchedEffect' we created that we might want to use here instead to prevent onPageView() being called again if there are any config changes like screen orientation?

feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/data/model/Notification.kt
```diff
@@ -0,0 +1,189 @@
...
+    @SerializedName("Status")
+    val status: String,
+    @SerializedName("DispatchedDateTime")
+    val rawDate: String,
+    @SerializedName("MessageTitle")
+    val messageTitle: String? = null,
+    @SerializedName("MessageBody")
+    val messageBody: String? = null,
+    @SerializedName("Metadata")
+    val metadata: Metadata
```
TimGJD, line 30: I think `Metadata` and ALL new classes in this PR which get serialised or deserialised using Gson will needing 'keeping' in a new proguard rules file? They fine on non-minify enabled builds but on minify enabled release builds we get exception when serialising and deserialising if not 'kept'.

TimGJD (approved):

PR #784, opened 2026-07-27
GOVUKAPP-3832 Driving data still showing

TimGJD:
# Driving data still showing

- Remove bottom nav bar from 'Your accounts' screen. 
- Complete DVLA unlinking function flow even when any parent coroutine jobs are cancelled. This is to meet a new requirement that, after confirming to unlink an account on 'Your accounts' screen, if users navigate away before the unlinking response has returned then the unlinking flow is still completed in the app.

## JIRA ticket(s)
  - [GOVUKAPP-3832](https://govukverify.atlassian.net/browse/GOVUKAPP-3832)

## Screen shots

| Before | After |
|--------|-------|
| Screenshot_1785158451 | Screenshot_1785158403 |

### DIFF
--- a/feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/di/DvlaModule.kt
+++ b/feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/di/DvlaModule.kt
@@ -65,6 +65,6 @@ object DvlaModule {
     @Singleton
     @Provides
     fun providesCoroutineScope(): CoroutineScope {
-        return CoroutineScope(SupervisorJob() + Dispatchers.Default)
+        return CoroutineScope(SupervisorJob() + Dispatchers.IO)
     }
 }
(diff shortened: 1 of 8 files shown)


feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/di/DvlaModule.kt
```diff
@@ -59,4 +62,9 @@ object DvlaModule {
         return StringProviderImpl(context)
     }
 
+    @Singleton
+    @Provides
+    fun providesCoroutineScope(): CoroutineScope {
```
lamprosgk: This would prevent another CoroutineScope provider to be added anywhere in the project (would cause a duplicate binding), the recommendation seems to be to create a custom `@Qualifier`

lamprosgk (approved):

PR #791, opened 2026-07-29
Integrate Qualtrics SDK v2

gclssvglx:
Qualtrics is a web- and app-based tool for survey creation, survey design and theming, survey delivery, and gathering survey results. 

It comes as a two-part system: 
  - eXperience Manager (XM) - a web-based tool that allows for the creation, styling and theming of feedback forms, surveys etc. It also gathers survey  responses and has some data analysis capability.
  - An SDK for app use, or JavaScript scripts for web use.

The SDK's job is to send events from the app to XM at appropriate points as the user is using the app. For simplicity, we are sending the same Firebase event - at the same point - to Qualtrics. This offers the analytics team the greatest range of potential target selectors, while respecting the user's analytics preferences.

XM evaluates these events against a number of user-defined triggers (built using the target selectors). 

The result of that evaluation is sent back to the app where the SDK uses it to show the survey, if appropriate. Survey's shown to the user can be ignored or taken. Either way the results of the survey are sent back to XM.

In order to use Qualtircs to is full potential, we need for the existing Firebase/Google analytics events to carry data that is also sent to Qualtrics - both via events and in the results - so they can be matched together to gain a full picture of the user's journey throughout the app and any survey/feedback they might give. 

To do this we use two key fields: 
  - `fb_session_id` - created when the app is booted or loaded and is unique until the next time it is booted or loaded.
  - `fb_user_pseudo_id` - also known as the app instance id, a unique id for the device (not the user).

## JIRA ticket(s)
  - [GOVUKAPP-3370](https://govukverify.atlassian.net/browse/GOVUKAPP-3370)
  - [GOVUKAPP-3782](https://govukverify.atlassian.net/browse/GOVUKAPP-3782)
  - [GOVUKAPP-3783](https://govukverify.atlassian.net/browse/GOVUKAPP-3783)

## Video 

Please note that the data in the Qualtrics eXperience Manager tool does take a short while arrive.

https://github.com/user-attachments/assets/a812ce07-73f3-4e21-9f8c-c05c5f0dad52

## Example Firebase/Qualtrics events

```shell
Logging event: 
  origin=app+gtm,
  name=Navigation,
  params=Bundle[{
    external=0, 
    fb_session_id=1784816277, 
    language=en, 
    ga_event_origin(_o)=app+gtm, 
    ga_screen_class(_sc)=HomeScreen, 
    ga_screen_id(_si)=-2738979775568515666, 
    ga_screen(_sn)=Homepage, 
    text=Settings, 
    type=Tab, 
    fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1
  }]

Logging event: 
  origin=app,
  name=screen_view(_vs),
  params=Bundle[{
    fb_session_id=1784816277, 
    language=en, 
    screen_title=Settings, 
    ga_event_origin(_o)=app, 
    engagement_time_msec(_et)=49933, 
    ga_previous_class(_pc)=HomeScreen, 
    ga_previous_id(_pi)=-2738979775568515666, 
    ga_previous_screen(_pn)=Homepage, 
    ga_screen_class(_sc)=SettingsScreen, 
    ga_screen_id(_si)=-2738979775568515665, 
    ga_screen(_sn)=Settings, 
    manual_tracking(_mst)=1, 
    fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1
  }]

Logging event: 
  origin=app+gtm,
  name=view_item_list,
  params=Bundle[{
    item_list_name=Your topics, 
    fb_session_id=1784816277, 
    item_list_id=Your topics, 
    ga_event_origin(_o)=app+gtm, 
    ga_screen_class(_sc)=HomeScreen, 
    ga_screen_id(_si)=-2738979775568515664, 
    ga_screen(_sn)=Homepage, 
    items=[Bundle[{
      item_name=Driving and transport, 
      item_list_name=Your topics, 
      item_list_id=Your topics,
      index=1, 
      location_id=driving-transport, 
      results=1
    }]], 
    fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1
  }]
```

## Example `qualtrics_survey_shown` and  `qualtrics_survey_closed` events

```shell
Logging event: 
  origin=app+gtm,
  name=qualtrics_survey_shown,
  params=Bundle[{
    fb_session_id=1784816277, 
    qualtrics_targeting_id=SI_8nRe24gWo6vCvBQ, 
    ga_event_origin(_o)=app+gtm, 
    ga_screen_class(_sc)=SettingsScreen, 
    ga_screen_id(_si)=-2738979775568515665, 
    ga_screen(_sn)=Settings, 
    fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1
  }]

Logging event: 
  origin=app+gtm,
  name=qualtrics_survey_closed,
  params=Bundle[{
    fb_session_id=1784816277, 
    qualtrics_targeting_id=SI_8nRe24gWo6vCvBQ, 
    ga_event_origin(_o)=app+gtm, 
    ga_screen_class(_sc)=SettingsScreen, 
    ga_screen_id(_si)=-2738979775568515665, 
    ga_screen(_sn)=Settings, 
    fb_user_pseudo_id=fe086b8a6caf295e4f6ece4a926622c1
  }]
```


analytics/src/main/kotlin/uk/gov/govuk/analytics/AnalyticsCoordinator.kt
```diff
@@ -0,0 +1,94 @@
...
+    private val qualtricsTargetingId = "qualtrics_targeting_id"
+    private val qualtricsSurveyShown = "qualtrics_survey_shown"
+    private val qualtricsSurveyClosed = "qualtrics_survey_closed"
+
+    override fun initialize() {
+        // Respect the user's privacy preference
+        if (analyticsRepo.analyticsEnabledState == AnalyticsEnabledState.ENABLED) {
+            qualtricsAnalyticsClient.initialize()
+            qualtricsAnalyticsClient.setOnSurveyClosedListener { targetingIds ->
+                onSurveyClosed(targetingIds)
```
davidc-gds, line 24: Do we need to check the analytics state before we send the survey closed event? Is it possible for the user to disable the analytics during a survey and would this result in an event being sent?

analytics/src/main/kotlin/uk/gov/govuk/analytics/AnalyticsCoordinator.kt
```diff
@@ -0,0 +1,94 @@
...
+            selectedItemIndex
+        )
+        qualtricsAnalyticsClient.logEcommerceEvent(
+            event,
+            ecommerceEvent,
+            onSurveyShown = { results -> onSurveyShown(results) }
+        )
+    }
+
+    private fun onSurveyShown(results: Map<String, TargetingResult>) {
```
davidc-gds, line 61: I'm not 100% on how qualtrics works but I think that logging an event triggers a re-evaluation which could in theory trigger a new survey being displayed. Is this something we need to worry about and test i.e. that showing/closing/completing a survey doesn't trigger a duplicate survey being displayed?

gclssvglx: Thanks @davidc-gds - this is an excellent spot! 

Yes, logically it is possible. However, it doesn't happen in the current setup for two reasons: 

1. We have a list of `analyticsParameterKeys` - all of these are blanked before each 'set' and 'send' to avoid a known bug in Qualtrics where it caches these values and resends them in subsequent events. This means that none of the values that the trigger is built on will appear in a `survey_shown` event.
2. There are rules set in Qualtrics around recontact and frequency - how often and who the survey will shown to. These have defaults (100 requests or 1 day), but can be configured. Doing so affects all surveys within a brand. It is unlikely that these would be set broad enough to make looping occur.

Having said that - if any of the triggers were broad enough (i.e. any event fires a survey) - then it would indeed keep asking the same user for feedback. It is my belief that analysts would never set such a trigger as doing so would quickly cause feedback-fatigue in our users.

analytics/src/main/kotlin/uk/gov/govuk/analytics/AnalyticsCoordinator.kt
gclssvglx: Thanks @davidc-gds - great spot. Yes, it is possible and yes it would result in an event being fired.

I've added a guard inside `onSurveyClosed` to stop the potential for this.

davidc-gds (dismissed):

davidc-gds (approved):

PR #793, opened 2026-07-30
GOVUKAPP-3850 Ability to link different DVLA account

TimGJD:
# Ability to link different DVLA account

**NOT DEV TESTED YET AS WAITING FOR BACKEND CHANGES BUT ASKED TO GET THROUGH REVIEW**

Despite the title/name this task is actually to prevent the ability to link a DVLA account to a different account by checking a session hash is the same session as the logged in users.

Added the following to run in this order:
- Receive new session hash param in verification response.
- Refresh tokens before starting DVLA auth flow.
- Send new session hash when starting DVLA auth flow.

## JIRA ticket(s)
  - [GOVUKAPP-3850](https://govukverify.atlassian.net/browse/GOVUKAPP-3850)

### DIFF
--- a/feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/DvlaViewModel.kt
+++ b/feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/DvlaViewModel.kt
@@ -18,6 +18,7 @@ import uk.gov.govuk.dvla.linking.data.LinkingRepo
 import uk.gov.govuk.dvla.navigation.ARG_DVLA_TOKEN
 import javax.inject.Inject
 import javax.inject.Named
+import androidx.core.net.toUri
 
 @HiltViewModel
 internal class DvlaViewModel @Inject constructor(
@@ -49,7 +50,10 @@ internal class DvlaViewModel @Inject constructor(
 
     sealed interface UiState {
         data object Default : UiState
-        data object Loading : UiState
+        sealed interface Loading : UiState {
+            data object Auth : Loading
+            data object Linking : Loading
+        }
         data object Success : UiState
         sealed interface Error : UiState {
             data object Offline : Error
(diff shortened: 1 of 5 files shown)


feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/DvlaViewModel.kt
```diff
@@ -78,22 +80,33 @@ internal class DvlaViewModel @Inject constructor(
...
+                result.value.verificationHash,
+                result.value.sessionHash
+            )
+
+            else -> _uiState.value = UiState.Error.Other
+        }
+    }
+
+    private fun launchAuthUrl(verificationHash: String, sessionHash: String) {
+        _authUrlToLaunch.value = "$dvlaAuthUrl?verification=$verificationHash&session=$sessionHash"
```
lamprosgk: Would be better to build the url with Uri builder, `Uri.parse(dvlaAuthUrl).build..`

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/DvlaViewModel.kt
```diff
@@ -78,22 +80,33 @@ internal class DvlaViewModel @Inject constructor(
...
     }
 
-    private fun startAuthFlow() {
+    private fun refreshTokens() {
         viewModelScope.launch {
-            when (val result = linkingRepo.getVerification()) {
-                is Result.Success -> launchAuthUrl(result.value.verificationHash)
-
-                else -> _uiState.value = UiState.Error.Other
+            if (authRepo.refreshTokens()) {
```
lamprosgk, line 94: Do we need any loading while we refresh the tokens?

TimGJD: It wasn't required but I've now added one which runs when we refresh tokens and get linking verification and session hashes. 

It's the existing linking loading screen just without any text:

Screenshot_1785424066

batesyuk (approved):

PR #794, opened 2026-07-30
Fix/not 417 delete sync issue

mattmapadmi:
Fixes two scenarios related to data races and the Notification list showing the wrong state:
* DELETE/PATCH request is slow and completes after the refresh of the notification list
* GET request completes before the server asynchronously processes the DELETE/PATCH

Also fixes a possible issue where the DELETE/PATCH request would be cancelled in-flightas it was tied to the ViewModel lifecycle, potentially creating a scenarion where the serverwould not process the full Request and update its state

Also fixes the Message count not updating in Settings as the load was only fired once (as it was driven by a function previously only intended for analytics).

### DIFF
--- a/feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/data/di/NotificationCentreModule.kt
+++ b/feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/data/di/NotificationCentreModule.kt
@@ -4,6 +4,9 @@ import dagger.Module
 import dagger.Provides
 import dagger.hilt.InstallIn
 import dagger.hilt.components.SingletonComponent
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
 import retrofit2.Retrofit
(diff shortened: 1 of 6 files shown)


feature/notificationcentre/src/main/kotlin/uk/gov/govuk/notificationcentre/data/di/NotificationCentreModule.kt
```diff
@@ -41,5 +44,10 @@ internal object NotificationCentreModule {
         return DateProviderImpl()
     }
 
+    @Provides
+    @Singleton
+    fun provideCoroutineScope(): CoroutineScope =
```
TimGJD, line 55: I think this would prevent another CoroutineScope provider to be added in the project? 
See line 30 in DvlaModule.kt we have added an internal custom qualifier for same reason, a resolution would be to do same here?

mattmapadmi: Ah I looked through the code for someone else providing a scope and completely missed that.

I'd seen the annotations in examples but thought they were a lot for just one usage, without realising Hilt can use them to differentiate (instead of using named instances either).

TimGJD (approved):

PR #803, opened 2026-08-04
GOVUKAPP-3822 Close menu navigation 

lamprosgk:
# Title

Added close menu option when talkback is on. `isTalkbackEnabled` function is copied from Chat and repeated, but felt like it didn't fit in any of the 'common' modules.

## JIRA ticket(s)
  - [GOVUKAPP-3822](https://govukverify.atlassian.net/browse/GOVUKAPP-3822)

## Screen shots

Screenshot_20260804_114628

### DIFF
--- a/feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/component/SummaryCardHeader.kt
+++ b/feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/component/SummaryCardHeader.kt
@@ -1,5 +1,7 @@
 package uk.gov.govuk.dvla.ui.component
 
+import android.content.Context
+import android.view.accessibility.AccessibilityManager
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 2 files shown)


feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/component/SummaryCardHeader.kt
```diff
@@ -93,29 +123,53 @@ private fun CardOverflowMenu(
...
                 )
             }
         }
     }
 }
+
+@Composable
+private fun OverflowMenuItemRow(
+    text: String,
+    altText: String? = null,
```
TimGJD: Could use our `AccessibleString` here instead?

feature/dvla/src/main/kotlin/uk/gov/govuk/dvla/ui/component/SummaryCardHeader.kt
```diff
@@ -93,29 +123,53 @@ private fun CardOverflowMenu(
...
+    onClick: () -> Unit
+) {
+    DropdownMenuItem(
+        text = {
+            BodyRegularLabel(
+                text = text,
+                color = GovUkTheme.colourScheme.textAndIcons.primary,
+                modifier = if (altText != null) {
```
TimGJD: Could use our Modifier extension `withAltText()` here instead?

lamprosgk: You're right, I just extracted the existing as it was 👍🏼

TimGJD (approved):

PR #813, opened 2026-08-05
GOVUKAPP-3524 Use BrowserLauncher everywhere

lamprosgk:
# Title

Added a new `:govkit` module for reusable code (existing `BrowserLauncher` could not be used in feature modules).
Moved play store and Gov.uk urls to `Urls` in govkit.
Added reusable `rememberPlayStoreLauncher()` and `rememberGovUkLauncher()` to launch play store listing and Gov.uk.

## JIRA ticket(s)
  - [GOVUKAPP-3524](https://govukverify.atlassian.net/browse/GOVUKAPP-3524)

## Figma
  - [Figma board]()

## Screen shots

| Before | After |
|--------|-------|
|        |       |


govkit/src/androidTest/java/uk/gov/govuk/govkit/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
+    @Test
+    fun useAppContext() {
+        // Context of the app under test.
+        val appContext = InstrumentationRegistry.getInstrumentation().targetContext
+        assertEquals("uk.gov.govuk.core.test", appContext.packageName)
```
TimGJD: If we ran the Android tests I think this would fail because package name is incorrect

lamprosgk: Good catch (had initially named core but renamed), will delete the example test, it's just boilerplate and doesn't add any value anyway.

govkit/src/main/kotlin/uk/gov/govuk/govkit/browser/BrowserLauncher.kt
```diff
@@ -70,13 +70,33 @@ internal sealed class BrowserActivityLauncher(
...
         }
     }
 }
 
+@Composable
+fun rememberPlayStoreLauncher(): () -> Unit {
+    val browserLauncher = rememberBrowserLauncher(shouldShowExternalBrowser = true)
+    return remember(browserLauncher) {
+        {
+            browserLauncher.launch(url = Urls.PLAY_STORE) { }
```
davidc-gds: Both `rememberPlayStoreLauncher()` and `rememberGovUkLauncher()` (see also line 95) pass an empty `onError = { }` to `launch`. If no app can handle the URL (`ActivityNotFoundException`), the tap silently does nothing — no toast, no alert. Elsewhere in the app (`GovUkApp.kt`) the same exception path sets `showBrowserNotFoundAlert = true`. Should these two follow the same pattern so the user gets feedback instead of a dead tap?

govkit/build.gradle.kts
```diff
@@ -0,0 +1,63 @@
...
+        property(
+            "sonar.coverage.exclusions",
+            properties["sonar.coverage.exclusions"].toString() + ",**/BrowserLauncher.*"
+        )
+    }
+}
+
+dependencies {
+    implementation(libs.androidx.browser)
+    implementation(libs.androidx.activity.compose)
```
davidc-gds, line 59: Other Compose-using modules (`design`, `feature/dvla`, etc.) pull in `implementation(platform(libs.androidx.compose.bom))` before declaring individual Compose deps, but this module doesn't. `androidx.activity.compose` here transitively brings in compose-runtime/compose-ui — worth pinning to the BOM so this module can't drift to a different Compose version than the rest of the app.

davidc-gds (commented):
- `app/build.gradle.kts:41` — the `PLAY_STORE_URL` buildConfigField looks unused now that `ForcedUpdateScreen`/`RecommendUpdateScreen` use `govkit`'s `Urls.PLAY_STORE` instead. Can this be removed?
- `BrowserLauncher.kt` — `rememberPlayStoreLauncher()` and `rememberGovUkLauncher()` are near-identical, differing only by the hardcoded URL, and each registers its own separate `ActivityResultLauncher`. Could these collapse into one generic `rememberUrlLauncher(url: String)`-style composable, similar to how `GovUkApp.kt` already calls `browserLauncher.launch(url) { onError }` with the URL supplied at call time? Not blocking, just flagging for maintainability as more destinations get added.

davidc-gds (approved):

PR #823, opened 2026-08-07
Qualtrics enhancement: allow survey authors to bypass prompt via query string parameter

gclssvglx:
This change allows Qualtrics survey authors to bypass a survey prompt by appending a query string parameter to the survey URL. 

This is configured in the Qualtrics eXperience Manager (XM) tool as follows...

### Inside the survey

Add the `hide_prompt` embedded data, so the survey knows that this field exists...

Screenshot 2026-08-07 at 15 12 03

Example URLs

- Embedded survey: `https://gdstrial.qualtrics.com/jfe/form/SV_3z0911Wln94GMse?Q_CHL=si_appsdk&fb_session_id=1786107706&fb_user_pseudo_id=c7d6111ee2b55858e9fbbf2144350203`
- Survey with prompt: `https://gdstrial.qualtrics.com/jfe/form/SV_3xyJtQMMpZprDYG?Q_CHL=si_appsdk&fb_user_pseudo_id=c7d6111ee2b55858e9fbbf2144350203&fb_session_id=1786107706`

### Inside the intercept

Add the `hide_prompt` static embedded data field and set it to `true`

Screenshot 2026-08-07 at 15 12 58

Example URLs

- Survey without prompt:  `https://gdstrial.qualtrics.com/jfe/form/SV_d6fDGXtifOAPV2u?Q_CHL=si_appsdk&fb_user_pseudo_id=c7d6111ee2b55858e9fbbf2144350203&fb_session_id=1786107706&hide_prompt=true`

### The result

Here the Edit Topics button fires an embedded survey, and the Chat screen view fires a survey with a prompt. However, a Settings screen view bypasses the prompt and takes the user directly to the survey itself.

[Screen_recording_20260807_155237.webm](https://github.com/user-attachments/assets/7293d086-7436-48e8-ae25-2e3639af37c9)

### DIFF
--- a/analytics/src/main/kotlin/uk/gov/govuk/analytics/QualtricsAnalyticsClient.kt
+++ b/analytics/src/main/kotlin/uk/gov/govuk/analytics/QualtricsAnalyticsClient.kt
@@ -96,29 +96,23 @@ class QualtricsAnalyticsClient @Inject constructor(
         qualtrics.registerViewVisit(eventName)
 
         qualtrics.evaluateProject { results ->
-            val targetResult = results.values.filterNotNull().lastOrNull {
-                !it.surveyUrl.isNullOrEmpty()
-            }
+            val passedResults = results.filterValues { it.passed() }
+
+            if (passedResults.isEmpty()) return@evaluateProject
+            val activity = activityProvider.currentActivity ?: return@evaluateProject
+
+            lastShownTargetingIds = passedResults.keys.toList()
+
+            val targetResult = passedResults.values.filterNotNull().firstOrNull()
             val surveyUrl = targetResult?.surveyUrl.orEmpty()
-            val skipPrompt = targetResult?.surveyUrl?.let {
-                it.toUri().query?.contains("hide_prompt=true")
-            } ?: false
+            val skipPrompt = targetResult?.surveyUrl?.toUri()?.query?.contains("hide_prompt=true") == true
 
-            val passedTargetingIds = results.filterValues {
-                it.passed()
-            }.keys.toList()
-
-            if (passedTargetingIds.isNotEmpty()) {
-                activityProvider.currentActivity?.let { activity ->
-                    lastShownTargetingIds = passedTargetingIds
-                    if (skipPrompt) {
-                        qualtrics.displayTarget(context, surveyUrl)
-                    } else {
-                        qualtrics.display(activity)
-                    }
-                    onSurveyShown?.invoke(results)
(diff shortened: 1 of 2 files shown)


davidc-gds:
One thing worth checking before this merges: targetResult (used for surveyUrl/skipPrompt) and passedTargetingIds (used for the "should we show anything" gate) are derived from two independent filters over results:

```
val targetResult = results.values.filterNotNull().lastOrNull {
    !it.surveyUrl.isNullOrEmpty()
}
...
val passedTargetingIds = results.filterValues {
    it.passed()
}.keys.toList()
```

targetResult is picked purely by "last entry with a non-empty surveyUrl" — it never checks it.passed(). So if results ever contains more than one entry (e.g. one targeting rule that passed but has no/different URL, and another rule that didn't pass but happens to carry a hide_prompt=true URL and sorts last), skipPrompt/surveyUrl could end up derived from a survey that never actually passed targeting for this event — and qualtrics.displayTarget(context, surveyUrl) would be called with the wrong survey's URL. There's no test covering a mixed passed/not-passed multi-entry results map, so this path isn't verified either way.

Checked with iOS — their logic is: take the first result that passed, and check whether that result's URL has hide_prompt. That both fixes the mismatch (URL selection is now scoped to passed()) and keeps platform parity. Suggested Android equivalent, filtering once and reusing it:

qualtrics.evaluateProject { results ->
    val passedResults = results.filterValues { it.passed() }
    if (passedResults.isEmpty()) return@evaluateProject
    val activity = activityProvider.currentActivity ?: return@evaluateProject

    lastShownTargetingIds = passedResults.keys.toList()

    val targetResult = passedResults.values.filterNotNull().firstOrNull()
    val surveyUrl = targetResult?.surveyUrl.orEmpty()
    val skipPrompt = targetResult?.surveyUrl?.toUri()?.query?.contains("hide_prompt=true") == true

    if (skipPrompt) {
        qualtrics.displayTarget(context, surveyUrl)
    } else {
        qualtrics.display(activity)
    }
    onSurveyShown?.invoke(results)
}

This also collapses the two separate filters into one (passedResults), so it's not possible for the URL check and the "did it pass" check to disagree.

gclssvglx:
https://github.com/govuk-once/govuk-mobile-android-app/pull/823#issuecomment-5280180975

Thanks @davidc-gds 👍 great spot! I've updated the code and added the missing tests.

davidc-gds (approved):

PR #829, opened 2026-08-13
GOVUKAPP-3827 Connected button group focus background

lamprosgk:
# Title

Added focus state for connected button (visible when using keyboard tab/arrows).

## JIRA ticket(s)
  - [GOVUKAPP-3827](https://govukverify.atlassian.net/browse/GOVUKAPP-3827)

## Figma
  - [Figma board](https://www.figma.com/design/x5i4a5RT0xMhJAcjoIxefb/2025-06-GOV.UK-app-library-v2--UX-refresh?node-id=14123-5629&t=pkxbBXO7OBvdE7wB-1)

## Screenshots

[Screen_recording_20260813_151129.webm](https://github.com/user-attachments/assets/60a94590-dcfa-4d8f-9bac-46b1d15a1ca0)

Screenshot_20260813_145403
Screenshot_20260813_145415
Screenshot_20260824_172548


design/src/main/kotlin/uk/gov/govuk/design/ui/component/Button.kt
```diff
@@ -233,29 +234,27 @@ fun ConnectedButton(
...
+    val isFocused by interactionSource.collectIsFocusedAsState()
 
-    val contentColour =
-        if (active) {
-            GovUkTheme.colourScheme.textAndIcons.header
-        } else {
-            GovUkTheme.colourScheme.textAndIcons.secondary
-        }
+    val (containerColour, contentColour) = when {
+        isFocused -> colours.containerFocused to GovUkTheme.colourScheme.textAndIcons.primary
```
davidc-gds: Will this not result in white text on a yellow background in dark mode? Could this potentially be an accessibility issue?

design/src/main/kotlin/uk/gov/govuk/design/ui/component/Button.kt
```diff
@@ -233,29 +234,27 @@ fun ConnectedButton(
...
     } else {
         text
     }
 
     Button(
         onClick = onClick,
-        modifier = modifier.clearAndSetSemantics {
+        interactionSource = interactionSource,
+        modifier = modifier
+            .focusable(interactionSource = interactionSource)
```
davidc-gds, line 256: Do we really need to add a focusable modifier here? Does the button (and interaction source we are passing in) not handle it? Could this result in the button receiving duplicate focus?

lamprosgk: The issue was that when talkback is on, the green 'ring' doesn't follow the key strokes and doesn't get applied to the button (see attached), that's the workaround I found

[Screen_recording_20260824_164926.webm](https://github.com/user-attachments/assets/39c7ef27-216b-4cd0-8c73-a990c0a2f82e)

design/src/main/kotlin/uk/gov/govuk/design/ui/component/Button.kt
lamprosgk: You're right, I'll fix that 👍🏼

davidc-gds (approved):

PR #842, opened 2026-08-28
GOVUKAPP-3953 : Suggested questions on chat screen

gclssvglx:
## Important

The full requirements are distributed between the JIRA ticket and the Figma.

## JIRA ticket(s)
  - [GOVUKAPP-3953](https://govukverify.atlassian.net/browse/GOVUKAPP-3953)

## Figma
  - [Figma board](https://www.figma.com/design/ogtppqfgxYnMiKA9OuGVCw/2026-GOV.UK-Chat?node-id=22752-20608&p=f&m=dev)

## Screen shots

| Light | Dark |
|--------|-------|
| Screenshot_20260902_112715 | Screenshot_20260902_112729 |
| Screenshot_20260902_114342 | Screenshot_20260902_114355 |
| Screenshot_20260902_112758 | Screenshot_20260902_112746 |
| Screenshot_20260902_114451 | Screenshot_20260902_114435 |

## Videos

[Screen_recording_20260902_133025.webm](https://github.com/user-attachments/assets/6e44e649-56cd-478f-8dee-8ec0a298b1c5)

[Screen_recording_20260902_133236.webm](https://github.com/user-attachments/assets/745fcaa7-0cec-4cf2-bd5d-c8787359a608)

https://github.com/user-attachments/assets/c006e22e-bc62-4ea0-b22c-e89074216ddc


feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/component/IntroMessages.kt
```diff
@@ -88,3 +130,89 @@ private fun Message(
         )
     }
 }
+
+@OptIn(ExperimentalLayoutApi::class)
```
davidc-gds, line 126: Should the example question visibility be tied to isLoading to prevent them being visible (and clickable) when we are waiting for an answer?

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/component/IntroMessages.kt
```diff
@@ -88,3 +130,89 @@ private fun Message(
...
+                Column(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .width(IntrinsicSize.Max)
+                        .padding(start = 60.dp)
+                        .semantics { isTraversalGroup = true },
+                    horizontalAlignment = Alignment.End
+
+                ) {
+                    chatExampleQuestions.forEachIndexed { index, question ->
```
davidc-gds: Loop var question is shadowing the function parameter question

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/component/IntroMessages.kt
```diff
@@ -55,10 +77,30 @@ internal fun IntroMessages(
...
+                        chatExampleQuestions
+                    )
+                }
             }
         } else {
-            Message()
+            Column {
+                Message()
+                MediumVerticalSpacer()
+                ExampleQuestions(
```
davidc-gds: If we have a conversation we don't want to show the example questions right. It performs that check inside the example questions function so when it enters this else branch it will emit nothing for the example questions. Can we just remove it from the else branch?

gclssvglx: Thanks @davidc-gds 👍 Yes, good spot only `message()` is needed thanks. Fixed up.

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/component/IntroMessages.kt
gclssvglx: Thanks @davidc-gds 👍 Yes, good spot thanks. Fixed up.

feature/chat/src/main/kotlin/uk/gov/govuk/chat/ui/component/IntroMessages.kt
gclssvglx: Thanks @davidc-gds 👍 I can't think of any instances when the example questions would be still visible when waiting for an answer - most actions to produce and submit a question should remove them. Having said that, it can't hurt to add that - just in case there's an edge-case or two that might mean it's possible. Fixed up.

davidc-gds (approved):

